PR #65, opened 2020-11-26
Simple client for vhive/knative and dockerized FunctionBench

ustiugov:
This PR introduces 

* simple clients to deploy and invoke functions for vHive/KNative. 
* OCI/docker images to use as functions (no need to review those files).
* logo and badges


knative_workloads/rnn_serving.yaml
```diff
@@ -1,7 +1,6 @@
 apiVersion: serving.knative.dev/v1 # Current version of Knative
 kind: Service
 metadata:
-  name: rnnserving # The name of the app
```
plamenmpetrov, line 4: why delete? does it even work like this?

knative_workloads/pyaes.yaml
```diff
@@ -1,7 +1,6 @@
 apiVersion: serving.knative.dev/v1 # Current version of Knative
 kind: Service
 metadata:
-  name: pyaes # The name of the app
```
plamenmpetrov, line 4: why delete? does it even work like this?

knative_workloads/helloworldSerial.yaml
```diff
@@ -1,7 +1,6 @@
 apiVersion: serving.knative.dev/v1 # Current version of Knative
 kind: Service
 metadata:
-  name: helloworldserial # The name of the app
```
plamenmpetrov, line 4: why delete? does it even work like this?

knative_workloads/helloworld.yaml
```diff
@@ -1,7 +1,6 @@
 apiVersion: serving.knative.dev/v1 # Current version of Knative
 kind: Service
 metadata:
-  name: helloworld # The name of the app
```
plamenmpetrov, line 4: why delete? does it even work like this?

examples/deployer/client.go
```diff
@@ -0,0 +1,157 @@
...
+
+	for i := 0; i < cap(sem); i++ {
+		sem <- true
+	}
+
+	return urls
+}
+
+func deployFunction(fID, filePath string) {
+	cmd := exec.Command(
```
plamenmpetrov, line 120: is there no go sdk we can use?

examples/deployer/client.go
```diff
@@ -0,0 +1,157 @@
...
+	)
+	stdoutStderr, err := cmd.CombinedOutput()
+	if err != nil {
+		log.Warnf("Failed to deploy function %s, %s: %v\n%s\n", fID, filePath, err, stdoutStderr)
+	}
+
+	log.Info("Deployed function", fID)
+}
+
+func writeURLs(filePath string, urls []string) {
```
plamenmpetrov, line 138: it would be good to specify which URL is for which function.

examples/invoker/client.go
```diff
@@ -0,0 +1,131 @@
...
+	pb "github.com/ustiugov/fccd-orchestrator/helloworld"
+	"google.golang.org/grpc"
+)
+
+var completed int64
+
+func main() {
+	urlFile := flag.String("urlFile", "urls.txt", "File with functions' URLs")
+	rps := flag.Int("rps", 1, "Target requests per second")
+	runTime := flag.Int("time", 5, "Run the benchmark for X seconds")
```
plamenmpetrov, line 44: runDuration?

examples/invoker/client.go
```diff
@@ -0,0 +1,131 @@
...
+
+	var lines []string
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		lines = append(lines, scanner.Text())
+	}
+	return lines, scanner.Err()
+}
+
+func runBenchmark(targetRPS, runTime int, urls []string) {
```
plamenmpetrov, line 73: order of arguments: urls, duration, RPS makes more sense to me

examples/invoker/client.go
```diff
@@ -0,0 +1,131 @@
...
+	atomic.AddInt64(&completed, 1)
+
+	return
+}
+
+func startMeasurement(msg string) (string, time.Time) {
+	return msg, time.Now()
+}
+
+func getDuration(msg string, start time.Time) {
```
plamenmpetrov, line 129: would be good to gather all of the latencies in one file to analyze/ccdf.

examples/invoker/client.go
```diff
@@ -0,0 +1,131 @@
...
+			url := urls[issued%len(urls)]
+			go invokeFunction(url)
+
+			issued++
+		}
+	}
+}
+
+func invokeFunction(url string) {
+	defer getDuration(startMeasurement(url)) // measure entire invocation time
```
plamenmpetrov, line 100: I feel like the URL should indicate which the function is or the URL file should have a mapping URL -> function. See comment in deployer

examples/deployer/client.go
```diff
@@ -0,0 +1,157 @@
...
+	var urls []string
+	sem := make(chan bool, deploymentConcurrency) // limit the number of parallel deployments
+
+	for k, fType := range funcSlice {
+		for i := 0; i < fType.Count; i++ {
+
+			sem <- true
+
+			fID := fmt.Sprintf("f%d-%d", k, i)
+			url := fmt.Sprintf("%s.%s", fID, "default.192.168.1.240.xip.io")
```
plamenmpetrov, line 99: make arguments like in cri/cri_test.go. This will be useful when deploying on other clusters

knative_workloads/rnn_serving.yaml
plamenmpetrov: never mind. i see why

ustiugov:
latencies are written to an output file but I haven't added the names of the corresponding functions. let's do it after the release (if it would be necessary).

examples/invoker/client.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func getDuration(msg string, start time.Time) {
+	latSlice.Lock()
+	defer latSlice.Unlock()
+
+	latency := time.Since(start).Microseconds()
+	log.Debugf("Invoked %v in %v usec\n", msg, latency)
+
+	latSlice.slice = append(latSlice.slice, latency)
```
plamenmpetrov, line 150: lock is only necessary for append. Please make sure the latency is measured before the lock is acquired or else the latency observed will include waiting for the lock

examples/invoker/client.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func writeLatencies(rps float64, latencyOutputFile string) {
```
plamenmpetrov, line 153: rps is not used in this function

PR #88, opened 2021-01-11
Subdirs testing disabled and CI and split into build and test.

shyamjesal:
- Split CI into build and test.

- Misc test moved to Github actions

- Taps test removed

Signed-off-by: Shyam Jesalpura <s.jesalpura@gmail.com>

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,54 +5,41 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ master ]
+  workflow_dispatch:
 
 env:
   GOOS: linux
   GO111MODULE: on
 
 jobs:
-
-  build:
-    strategy:
-      matrix:
-        go: ['1.13', '1.14', '1.15']
-      # Build all variants regardless of failures
-      fail-fast: false
-
-    name: Build and check code quality
+  test:
+    name: Integration tests on Github Runners
     runs-on: ubuntu-18.04
+    strategy:
+      fail-fast: false
+      matrix:
+        module: [taps, misc]
     steps:
-
-    - name: Set up Go ${{ matrix.go }}
+    - name: Set up Go 1.x
       uses: actions/setup-go@v2
       with:
-        go-version: ${{ matrix.go }}
+        go-version: 1.14
 
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
-      with:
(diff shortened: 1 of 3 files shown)


ustiugov:
PR name is misleading. 

Please add the following clean-up at the end of each test in yaml that engages containerd:
```
    - name: Cleaning
      if: ${{ always() }}
      run: ./scripts/clean_fcctr.sh
```

So that if a test run fails, it would not break all further runs.

function-images/lr_serving/Dockerfile
```diff
@@ -13,6 +13,7 @@ RUN apk add --no-cache \
     pip install --disable-pip-version-check --no-build-isolation pandas==1.0.5 && \
     pip install --disable-pip-version-check --no-build-isolation protobuf==3.11.3 && \
     pip install --disable-pip-version-check --no-build-isolation grpcio==1.26.0 && \
+    pip install --disable-pip-version-check --no-build-isolation pybind11 && \
```
ustiugov, line 16: can you please explain the fix?

function-images/pyaes/Dockerfile
```diff
@@ -5,6 +5,7 @@ RUN pip3 install --user -r requirements.txt
 
 FROM vhiveease/py_grpc:base as var_workload
 COPY *.py /
+RUN true
```
ustiugov, line 8: can you please explain the fix here and elsewhere?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,65 @@
...
+
+    - name: Build
+      run: go build -race -v -a ./...
+
+    - name: Run tests in submodules
+      run: |
+        make -C ctriface test
+        make -C ctriface test-man
+
+    - name: Run vHive end-to-end tests
```
ustiugov, line 61: the following two jobs must separate from the subdirs job, right?

ustiugov (changes requested):
please do add the cleaning script with the next commit.

function-images/lr_serving/Dockerfile
shyamjesal: [Scipy/fft](https://github.com/scipy/scipy/blob/v1.6.0/scipy/fft/_pocketfft/setup.py) build failed due to missing pybind11 dependency.

PR #177, opened 2021-03-17
Add Knative eventing in default cluster setup

shyamjesal:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Zipkin support added for tracing Knative function call requests. More info [here](./docs/developers_guide.md#Knative-request-tracing)
 - added support for MinIO object store. More info [here](./docs/developers_guide.md#MinIO-S3-service)
 - Added an automated tail-latency-aware profiler that collects the metrics for [TopDown](https://ieeexplore.ieee.org/document/6844459) characterization from Intel.
+- [alpha] Added knative eventing support using In-Memory Channel and MT-Channel-broker. Integration tests missing and Apache Kafka support coming soon.
 
 ### Changed
 
--- a/scripts/cluster/setup_master_node.sh
+++ b/scripts/cluster/setup_master_node.sh
@@ -60,4 +60,15 @@ fi
 kubectl apply --filename $ROOT/configs/knative_yamls/serving-default-domain.yaml
 
 kubectl apply --filename https://github.com/knative/net-istio/releases/download/v0.19.0/release.yaml
+
+# install knative eventing
+kubectl apply --filename https://github.com/knative/eventing/releases/download/v0.21.0/eventing-crds.yaml
+kubectl apply --filename https://github.com/knative/eventing/releases/download/v0.21.0/eventing-core.yaml
+
+# install a default Channel (messaging) layer
+kubectl apply --filename https://github.com/knative/eventing/releases/download/v0.21.0/in-memory-channel.yaml
+
+# install a Broker (eventing) layer:
+kubectl apply --filename https://github.com/knative/eventing/releases/download/v0.21.0/mt-channel-broker.yaml
+
 kubectl --namespace istio-system get service istio-ingressgateway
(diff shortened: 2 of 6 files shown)


scripts/cluster/setup_master_node.sh
```diff
@@ -60,4 +60,15 @@ fi
 kubectl apply --filename $ROOT/configs/knative_yamls/serving-default-domain.yaml
 
 kubectl apply --filename https://github.com/knative/net-istio/releases/download/v0.19.0/release.yaml
-kubectl --namespace istio-system get service istio-ingressgateway
+
+# install knative eventing
+kubectl apply --filename https://github.com/knative/eventing/releases/download/v0.21.0/eventing-crds.yaml
```
ustiugov, line 65: please make v0.21.0 a variable that is used for the stock serving and eventing.

scripts/cluster/setup_master_node.sh
```diff
@@ -60,4 +60,15 @@ fi
...
+kubectl apply --filename https://github.com/knative/eventing/releases/download/v0.21.0/eventing-core.yaml
+
+# install a default Channel (messaging) layer
+kubectl apply --filename https://github.com/knative/eventing/releases/download/v0.21.0/in-memory-channel.yaml
```
ustiugov, line 69: todo: need to replace this with Kafka

scripts/cluster/setup_master_node.sh
shyamjesal: done

scripts/cluster/setup_master_node.sh
shyamjesal: added as a comment.

scripts/cluster/setup_master_node.sh
```diff
@@ -60,4 +60,17 @@ fi
 kubectl apply --filename $ROOT/configs/knative_yamls/serving-default-domain.yaml
 
 kubectl apply --filename https://github.com/knative/net-istio/releases/download/v0.19.0/release.yaml
-kubectl --namespace istio-system get service istio-ingressgateway
+
```
ustiugov, line 63: we should have the same version for serving and eventing since Knative folks seem to have them in sync. please use one variable for both.

scripts/setup_zipkin.sh
```diff
@@ -22,9 +22,12 @@
...
 
+set -e
 # install zipkin pods
 kubectl apply -f https://raw.githubusercontent.com/istio/istio/release-1.9/samples/addons/extras/zipkin.yaml
 
+sleep 10s
+
 # enable tracing in Knative
 kubectl patch configmap/config-tracing \
   -n knative-serving \
```
ustiugov, line 33: I guess this is the name of the namespace, right? probably, you were missing info in the traces before because some of the services were running in different namespaces. What do you think?

ustiugov (changes requested):

scripts/setup_zipkin.sh
shyamjesal: this namespace is for the config file location. Namespaces for the zipki pods are set by the config file provided by the istio team (line number 27).

scripts/cluster/setup_master_node.sh
shyamjesal: done

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - Zipkin support added for tracing Knative function call requests. More info [here](./docs/developers_guide.md#Knative-request-tracing)
 - added support for MinIO object store. More info [here](./docs/developers_guide.md#MinIO-S3-service)
 - Added an automated tail-latency-aware profiler that collects the metrics for [TopDown](https://ieeexplore.ieee.org/document/6844459) characterization from Intel.
+- [alpha] Added knative eventing support.
```
ustiugov, line 12: Please specify that only their in-memory channel is supported as of now and through integration tests are missing (unless you plan to add an eventing example in this PR). Maybe adding an invocation without validation in this PR is a good idea if it is more or less trivial. If it would take more than 30min, let's postpone it to another PR.

ustiugov (approved):

PR #185, opened 2021-03-19
fetch latest binaries before running CI tests

shyamjesal:

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -35,9 +35,20 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: '3.x'
+    
+    - name: Upgrade git
+      run: |
+        sudo add-apt-repository ppa:git-core/ppa -y
+        sudo apt update
+        sudo apt install git -y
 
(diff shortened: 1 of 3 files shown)


.github/workflows/integration_tests.yml
```diff
@@ -35,9 +35,20 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: '3.x'
+    
+    - name: Check git version
```
ustiugov, line 39: it's not a check but installation

ustiugov:
what about knative containers, btw? are they pulled on every CRI test?

.github/workflows/integration_tests.yml
```diff
@@ -35,9 +35,20 @@ jobs:
...
+      run: |
+        sudo add-apt-repository ppa:git-core/ppa -y
```
ustiugov, line 41: this needs to be done as a part of the runner's dockerfile.

ustiugov (changes requested):

ustiugov (dismissed):

ustiugov (approved):

PR #186, opened 2021-03-19
add stack charts plotter for metric groups

NiuJ1ao:
- plotter detects any metric groups in the .csv file and plot stack charts for them.

For example, if a CSV file has _Frontend_Bound_, _Bad_Speculation_, _Backend_Bound_ and _Retiring_, The plotter plots a stack chart for _Level_1_ (level 1 is the parent of four nodes) as following:
Level_1

Signed-off-by: NiuJ1ao <26167136+NiuJ1ao@users.noreply.github.com>


profile/toplev_metrics.json
```diff
@@ -0,0 +1,107 @@
+{"Level_1":
```
ustiugov, line 1: if this file is copied from the pmu tools repo, it should be fetched at the setup time and should not be in this repo.

profile/plotter.go
```diff
@@ -34,17 +38,25 @@ import (
...
-func PlotCVS(xStep int, filePath, inFile, xLable string) {
+// PlotLineCharts plots every attribute as VM number increases in the csv file
+func PlotLineCharts(xStep int, filePath, inFile, xLable string) {
 	var (
 		records = readResultCSV(filePath, inFile)
 		rows    = len(records)
 		cols    = len(records[0])
 	)
 
+	if rows < 3 {
```
ustiugov, line 55: why 3? add a comment

ustiugov (changes requested):
please add a thorough description of the feature with an example of a plot. The code seems fine apart from few comments. the json should be fetched at the setup time unless there is a good reason for having this file in this repo.

profile/toplev_metrics.json
NiuJ1ao: This file is not from pmu tools repo. JSON structure would be easier to parse and check the relations between nodes.

profile/plotter.go
NiuJ1ao: Done

ustiugov:
you should not put the plot into the repo unless you add it to `*md` docs (which you may want to do). For the PR's description, you can just drag-and-drop a figure/screenshot when editing.

ustiugov:
please merge the last 2 commits. please try to avoid creating commits with the same logical value.

ustiugov (changes requested):
please merge the last 2 commits

ustiugov (changes requested):
please remove the figure from the docs/ folder

ustiugov (dismissed):

ustiugov (approved):

Issue #165, opened 2021-03-11
Deploy and populate a local container registry


ustiugov:
For the FaaS benchmarking at scale, we would like to avoid the obvious bottleneck of a remote registry, like DockerHub. 

Required changes:

- [ ] the ctriface module should not default to `docker.io/`
- [ ] YAML files should specify the registry when specifying each container image
- [ ] need to add a script that would deploy by URL and populate the local registry in a vHive cluster, by copying the images from remote registries, e.g., DockerHub, to the cluster-local registry
- [ ] tests need to be updated accordingly

We welcome a contribution from @amohoste on this issue.

amohoste:
How would you like the local registry population to work? Do you want it to be automated such that upon calling the deployer, images are first pulled to the local registry or did you have a separate script in mind. Also, there should be a way for the script to know which global registry it can fetch the images from. A straightforward way to do this would be to map the local url to the global url of a specified registry. For example if a user enters `docker-registry.registry.svc.cluster.local:5000/vhiveease/pyaes:var_workload` in the yaml file, the script tries to pull the image from `docker.io/vhiveease/pyaes:var_workload`.

Please feel free to give other suggestions.

ustiugov:
@amohoste sorry for the late response. 

I would prefer to give complete flexibility of using different registries, incl. dockerhub, AWS ecr, vHive local registry, etc.,  to the user who specifies the YAML files. The script can take a list of images and two arguments: source registry and destination registry.

amohoste:
Okay, I suppose it should be in a separate script it is then?

Regarding the testcases, should I make them point to docker.io and add some for the local registry or should I port all of them to the local registry?

ustiugov:
yes, separate. In the unit tests as well as the integration tests except in the `cri/` module, you can use `docker.io`. In `cri`, we test integration with kubernetes so there it would be good to use the local registry instead (e.g., for one of the functions). The CRI tests are not 100% stable as of now so please let us know in case of any weird failures.

amohoste:
Could you point me to where the test images for the CRI tests get deployed? `cri_test.go` is calling the following hardcoded url: `helloworld.default.192.168.1.240.xip.io:80` but I'm not 100% sure where this is getting deployed exactly. If we would want to test a function deployed from the local registry, I think we also would also need to deploy a different image than helloworld since we can't have 2 images with the same name.

ustiugov:
@amohoste these are three different entities. 
[...]

PR #188, opened 2021-03-20
Deployment of a local registry

amohoste:
Implementation of a local registry as specified in #165. 

* Skopeo is installed on node setup, allowing images to be transferred between registries without docker
* A local registry is deployed upon setup of the master node
* A script has been added to populate the local registry using images from a remote registry
* Test have been updated to include an image from the local registry

**Notes**
* Skopeo installation without compiling from source only works for ubuntu 18.04+ or Debian 10
* The local registry does not support SSL
* When no registry is specified in the imagename, the orchestrator defaults to docker.io, which is the default Kubernetes behavior
* Storage size for the local registry volume has been set to 5 GiB for now. This might need to be adjusted depending on the requirements


examples/registry/go.mod
```diff
@@ -0,0 +1,5 @@
+module deployer
+
+go 1.16
```
ustiugov, line 3: please set to 1.15 like in other modules.

scripts/setup_system.sh
```diff
@@ -39,6 +39,19 @@ sudo apt-get -y install \
     gnupg-agent \
     software-properties-common >> /dev/null
 
+# Install Skopeo (ubuntu 18.04+ or Debian 10)
```
ustiugov, line 42: for simplicity, please assume ubuntu 18. Please add the repo to the sources list before line 25.

scripts/setup_system.sh
```diff
@@ -39,6 +39,19 @@ sudo apt-get -y install \
...
+if [ "$(grep -Ei 'buntu' /etc/os-release)" ]; then
+   . /etc/os-release
+    echo "deb https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_${VERSION_ID}/ /" | sudo tee /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list
+    curl -L https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_${VERSION_ID}/Release.key | sudo apt-key add -
+else
+    echo 'deb https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/Debian_10/ /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list
+    curl -L https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/Debian_10/Release.key | sudo apt-key add -
+fi
+sudo apt-get update
+sudo apt-get -y upgrade
```
ustiugov, line 52:
```suggestion
```

scripts/setup_system.sh
```diff
@@ -39,6 +39,19 @@ sudo apt-get -y install \
...
+sudo apt-get -y install skopeo
```
ustiugov, line 53: move to lines 27-40

examples/registry/populate_registry.go
```diff
@@ -0,0 +1,109 @@
+// MIT License
+//
+// Copyright (c) 2020 Dmitrii Ustiugov and EASE lab
```
ustiugov, line 3: you may want to put your name here + "and EASE lab"

examples/registry/go.mod
```diff
@@ -0,0 +1,5 @@
+module deployer
```
ustiugov, line 1: I don't think that you need a Go module for a simple script.

ctriface/iface.go
```diff
@@ -272,15 +276,65 @@ func (o *Orchestrator) StopSingleVM(ctx context.Context, vmID string) error {
 	return nil
 }
 
+func isLocalDomain(s string) (bool, error) {
```
ustiugov, line 279: please add a comment about this function

configs/registry/repository-volume.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: docker-repo-pv
+  labels:
+    usage: registry
+spec:
+  capacity:
+    storage: 5Gi # TODO: size
```
ustiugov, line 9: as for lines 9 and 25 (i.e., storage size), please use an environment variable from the setup_system.sh script. 5GB is a fine constant to me. 
Please take a look at how we set up MinIO storage size dynamically [here](https://github.com/ease-lab/vhive/blob/master/docs/developers_guide.md#minio-s3-service)

ustiugov (changes requested):
good job! the code looks generally good. please see my, mostly minor, comments inline.

amohoste:
Thanks for the review, I'll also try to run the tests locally to see if anything is wrong on my end. Regarding the spellcheck, it is complaining about a command line argument used in a markdown file. Should I add that parameter to the exceptions?

ustiugov:
yes wrt the spellcheck, don't worry about the test failures as of now.

scripts/setup_system.sh
amohoste: Okay, so you mean we drop support for Debian? In that case we should adjust the quickstart guide to only mention Ubuntu.

ustiugov: we say that we tested on Ubuntu 18, I think that it's enough. We should keep the setup simple.

examples/registry/populate_registry.go
```diff
@@ -0,0 +1,109 @@
...
+		log.Fatalf("failed to build url: %v", err)
+	}
+
+	destURL, err := buildURL(image, destRepo)
+	if err != nil {
+		log.Fatalf("failed to build url: %v", err)
+	}
+
+	cmd := exec.Command(
+		"skopeo",
```
ustiugov, line 99: you need to use the absolute path here. you can see the warning [here](https://github.com/ease-lab/vhive/pull/188/checks?check_run_id=2166170737#step:7:417)

examples/registry/populate_registry.go
```diff
@@ -0,0 +1,109 @@
...
+		"copy",
+		sourceURL,
+		destURL,
+		"--dest-tls-verify=false",
+	)
+	stdoutStderr, err := cmd.CombinedOutput()
+	if err != nil {
+		log.Warnf("Failed to pull image %s: %v\n%s\n", image, err, stdoutStderr)
```
ustiugov, line 107: this must be an error (i.e., fatal) instead of a warning

ustiugov (changes requested):

examples/registry/populate_registry.go
amohoste: Could this be because setup_system.sh is not called in the CRI github workflow and thus not installed? Skopeo should have been added to the path through apt once the setup is done.

ustiugov (approved):

ustiugov (approved):

PR #189, opened 2021-03-22
update knative serving to v0.21

shyamjesal:


ustiugov:
need to rebase

scripts/setup_system.sh
```diff
@@ -35,6 +35,7 @@ sudo apt-get -y install \
     net-tools \
     git-lfs \
     bc \
```
ustiugov, line 37: i think you should add gettext-base here as well for the users that would use a local registry on bare metal.

shyamjesal: added.

PR #190, opened 2021-03-22
Add gettext dependency

amohoste:
Gettext was not installed causing [the master node setup to fail](https://github.com/ease-lab/vhive/runs/2169277370?check_suite_focus=true#step:7:273) during the tests.


ustiugov:
fixed in #189

PR #193, opened 2021-03-23
Update firecracker/jailer to v0.24-firecracker (with REAP support)

ustiugov:
- [x]  update firecracker and jailer binaries to v0.24 with user page faults

Signed-off-by: Dmitrii Ustiugov <dmitrii.ustiugov@epfl.ch>


ustiugov:
I've validated that performance of the vanilla and the REAP snapshots is not significantly different than in with firecracker v0.21 that we used for ASPLOS'21 paper evaluation. I suggest we merge this PR and update firecracker-containerd binaries later on. I don't anticipate those changes to be important so they can be done off the critical path.

The binary is built upon firecracker v0.24 with REAP support ([branch](https://github.com/ease-lab/firecracker/tree/v0.24_user_page_faults)).

shyamjesal (approved):

PR #216, opened 2021-04-18
Refactor CI env variables

ustiugov:
* move duplicated env variable declarations to the workflow and job levels
* add job_id to nightly CRI logs to avoid parallel jobs overwrite each other's artifacts

Signed-off-by: Dmitrii Ustiugov <dmitrii.ustiugov@epfl.ch>

### DIFF
--- a/.github/workflows/nightly_tests.yml
+++ b/.github/workflows/nightly_tests.yml
@@ -8,12 +8,19 @@ on:
 env:
   GOOS: linux
   GO111MODULE: on
+  TMPDIR: /root/tmp/
+  GOCACHE: /root/tmp/gocache
+  GOPATH: /root/tmp/gopath
+  GOROOT: $HOME/go
 
 jobs:
   integration-tests:
     name: Test all functions
     runs-on: [self-hosted, nightly, integ]
     steps:
+    - name: Setup TMPDIR
+      run: mkdir -p $TMPDIR
+
     - name: Set up Go 1.15
       uses: actions/setup-go@v2
       with:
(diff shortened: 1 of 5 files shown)


.github/workflows/nightly_tests.yml
```diff
@@ -99,7 +96,7 @@ jobs:
       if: ${{ always() }}
       uses: actions/upload-artifact@v2
       with:
-        name: ctrd-logs
+        name: ${{ github.job }}-${{ matrix.vhive_args }}-ctrd-logs
```
shyamjesal, line 99: Why use github job here as the log directory uses only `$GITHUB_RUN_ID` as the identifier?

shyamjesal:
Shouldn't we modify the `CTRDLOGDIR` variable inside `start_onenode_vhive_cluster.sh` in order to avoid log rewrites by parallel jobs?

.github/workflows/nightly_tests.yml
shyamjesal, line 99: Is changing the name here enough?
Value of `CTRDLOGDIR` inside `start_onenode_vhive_cluster.sh` also will have to be changed, right?

shyamjesal (changes requested):
Shouldn't we modify the CTRDLOGDIR variable inside start_onenode_vhive_cluster.sh in order to avoid log rewrites by parallel jobs?

.github/workflows/nightly_tests.yml
ustiugov: the folder name can be the same because one runner processes one job at a time. However, each folder must be uploaded as a separate archive to be available for downloading.

shyamjesal (approved):

shyamjesal (approved):

PR #247, opened 2021-05-21
Quieter setup_node.sh

boramalper:
Make `setup_node.sh` quieter and more compliant

- Use "quiet" flag on commands that do not need to be verbose
- Prefer commands intended for scripting usage (e.g., `apt-get` over `apt`)

Edit: Just realised that I built this upon my previous work on updating the quickstart guide so this PR includes that commit too. Sorry! Not sure how to isolate this and that, but let me know if I can help.

### DIFF
--- a/scripts/install_go.sh
+++ b/scripts/install_go.sh
@@ -24,7 +24,7 @@
 
 set -e
 
-wget https://golang.org/dl/go1.15.8.linux-amd64.tar.gz
+wget --continue --quiet https://golang.org/dl/go1.15.8.linux-amd64.tar.gz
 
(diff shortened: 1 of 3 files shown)


ustiugov:
good job with the bash scripts, let us merge this change before the documentation. 

You need to remove the unrelated commits from this PR. Probably, it would be easier to create a new branch from the tip of the main branch, adding a single commit with the bash scripts' changes.

boramalper:
Thank you. I reverted the quickstart guide commit, hope this works!

boramalper:
@ustiugov Now there is a single commit.

ustiugov (approved):

PR #268, opened 2021-06-07
Create single module, aggregate dependencies

### LINKED ISSUE #263: Create single module, aggregate dependencies
Currently, we have many modules, for which we have `go.mod`s, where the dependencies for each module are kept. This make upgrading some dependencies such as firecracker-containerd error prone, because the dependency needs to be changed in several places.

It seems a lot of the dependencies overlap across the modules - containerd, firecracker-containerd, and some less obvious ones, like grpc. This is one reason why I propose to keep the non-top modules as packages - `misc`, `taps`, etc. Another reason is that really these modules are not meant to be used outside of vHive in order to warrant keeping them as modules if someone wants to use them.

I propose the following way to do this:
1. Delete `go.mod` and `go.sum` in all modules we want to keep as packages. Import paths should stay the same.
2. Introduce the missing dependencies with version to the top-level `go.mod`.

boramalper:
Fixes #263.

I'll rebase them into a single commit, left as-is for clarity for now.


ctriface/iface_test.go
```diff
@@ -40,7 +40,7 @@ import (
 var (
 	isUPFEnabled = flag.Bool("upf", false, "Set UPF enabled")
 	isLazyMode   = flag.Bool("lazy", false, "Set lazy serving on or off")
-	isWithCache  = flag.Bool("withCache", false, "Do not drop the cache before measurements")
+	// isWithCache  = flag.Bool("withCache", false, "Do not drop the cache before measurements")
```
ustiugov, line 43: we need this option

memory/manager/manager_test.go
```diff
@@ -20,6 +20,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
+//nolint:unused
```
ustiugov, line 23: I agree with this workaround, however please raise an Issue

boramalper: Done, see #270.

ustiugov (changes requested):
Please make the commits' comments slightly more verbose. it is not clear what you do with each package from commit messages.

Rest is good, thanks.

ustiugov:
please rebase

boramalper:
@ustiugov 2 tests are failing but otherwise lgtm, care to merge?

ustiugov (changes requested):
More minor things:

- [x] add a not to Changed in the changelog (simplified go dependencies management in a single file)
- [x] remove the corresponding module from dependabot's config: https://github.com/ease-lab/vhive/blob/main/.github/dependabot.yml

ustiugov (approved):

PR #295, opened 2021-06-23
Enabled Istio tracing

ustiugov:
Need to check if Istio ingress appears in Zipkin. May require further changes.


ustiugov:
@MBaczun can you please check?

PR #310, opened 2021-06-29
Setup gvisor-containerd service

alexandrinapanfil:
What this patch does: Sets up `gvisor-containerd`, a stock `containerd` service that has separate configuration, socket address and shim to the `runsc` low-level runtime. 
This service will work separately to the other two existing `containerd` services and it will run gVisor-sandboxes.

### DIFF
--- a/scripts/install_stock.sh
+++ b/scripts/install_stock.sh
@@ -36,6 +36,10 @@ wget --continue --quiet https://github.com/opencontainers/runc/releases/download
 mv runc.amd64 runc
 sudo install -D -m0755 runc /usr/local/sbin/runc
 
+wget --continue --quiet https://storage.googleapis.com/gvisor/releases/release/20210622/x86_64/runsc 
+sudo chmod a+rx runsc
+sudo mv runsc /usr/local/bin
+
 containerd --version || echo "failed to build containerd"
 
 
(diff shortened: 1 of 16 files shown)


scripts/install_stock.sh
```diff
@@ -36,6 +36,10 @@ wget --continue --quiet https://github.com/opencontainers/runc/releases/download
 mv runc.amd64 runc
 sudo install -D -m0755 runc /usr/local/sbin/runc
 
+wget --continue --quiet https://storage.googleapis.com/gvisor/releases/release/latest/x86_64/runsc
```
ustiugov, line 39: let's fix the release to a concrete version that is currently the latest stable one. this would allow to avoid incompatibility problems that might come with new versions.

ustiugov (changes requested):
Minor changes

ustiugov:
to enable CI, please add feature/gvisor branch to the [workflow files](https://github.com/ease-lab/vhive/tree/main/.github/workflows), as follows:

```
on:
  push:
    branches: [ main,  feature/gvisor ]
  pull_request:
    branches: [ main, feature/gvisor ]
```

scripts/install_stock.sh
alexandrinapanfil: Changed runsc release to a fixed one.

PR #365, opened 2021-08-01
Stacking training benchmark

ustiugov:
This benchmark is representative of production serverless training workloads, examples:

- https://www.mdpi.com/1996-1073/11/4/949/htm
- https://www.hindawi.com/journals/scn/2020/4586875/
- https://link.springer.com/article/10.1007/s10346-019-01286-5

The benchmark has 4 functions:
- **driver** orchestrates the entire flow. Starts by uploading the dataset for the trainers and the metatrainer, collects the final models.
- a set of **trainers** that fit a model each (tested with 4 and 16 trainers, sequentially and in parallel)
- **reducer** collects the models and predictions from each trainer.
- **metatrainer** trains together with the trained models' layer, finalizing the 2-layer model.

Currently, only S3 is supported, tested with docker-compose only.


function-images/tests/stacking-training/Makefile
```diff
@@ -0,0 +1,88 @@
...
+
+# For debugging on Apple M1 below
+#PLATFORM = --platform linux/arm64
+#MODE = --push
+
+.PHONY: proto
+
+all: all-image 
+
+all-image: driver-image trainer-image reducer-image metatrainer-imag
```
MBaczun, line 37:
```suggestion
all-image: driver-image trainer-image reducer-image metatrainer-image
```

MBaczun:
In the python code the `MAX_SERVER_THREADS` env var is sometimes used, but in the docker-compose and now also in the kn manifests we set `MAX_RECOG_SERVER_THREADS`. Looks like a mistake to me, prob needs fixing?

function-images/tests/stacking-training/Makefile
ustiugov: yep please fix

ustiugov:
good catch! let's use `MAX_SERVER_THREADS` everywhere in this benchmark

> In the python code the `MAX_SERVER_THREADS` env var is sometimes used, but in the docker-compose and now also in the kn manifests we set `MAX_RECOG_SERVER_THREADS`. Looks like a mistake to me, prob needs fixing?

ustiugov:
try to re-generate the proto files, the build test fail and it's weird. it might be due to the fact that I generated them on my Apple M1 (arm) laptop....

function-images/tests/stacking-training/driver/main.py
```diff
@@ -0,0 +1,350 @@
...
+
+    def put(self, obj, key):
+        msg = "Driver uploading object with key '" + key + "' to " + self.transferType
+        log.info(msg)
+        with tracing.Span(msg):
+            if self.transferType == S3:
+                self.put_to_s3(obj, key)
+            elif self.transferType == XDT:
+                log.fatal("XDT is not supported")
+        return
```
ustiugov, line 207:
```suggestion
        return key
```

function-images/tests/stacking-training/driver/main.py
```diff
@@ -0,0 +1,350 @@
...
+
+    def put_to_s3(self, obj, key):
+        s3object = self.s3_client.Object(bucket_name=self.benchName, key=key)
+        pickled = pickle.dumps(obj)
+        s3object.put(Body=pickled)
+
+    def get_from_s3(self, key):
+        obj = self.s3_client.Object(bucket_name=self.benchName, key=key)
+        response = obj.get()
+        return pickle.loads(response['Body'].read())
```
ustiugov, line 226: let's move pickle.loads and pickle.dumps to get() and put() because it would be necessary for all S3, XDT, Elasticache

function-images/tests/stacking-training/manifests/service-trainer.yaml
```diff
@@ -0,0 +1,56 @@
...
+metadata:
+  name: trainer
+  namespace: default
+spec:
+  template:
+    metadata:
+      annotations:
+        # prevent cold boots
+        autoscaling.knative.dev/minScale: "1"
+        autoscaling.knative.dev/maxScale: "1"
```
ustiugov, line 34:
```suggestion
        autoscaling.knative.dev/minScale: "4"
        autoscaling.knative.dev/maxScale: "4"
```

should be equal to numTrainers parameter in Driver

function-images/tests/stacking-training/manifests/service-trainer.yaml
```diff
@@ -0,0 +1,56 @@
...
+          env:
+            - name: TRANSFER_TYPE
+              value: "S3"
+            - name: AWS_ACCESS_KEY
+              value: ${AWS_ACCESS_KEY}
+            - name: AWS_SECRET_KEY
+              value: ${AWS_SECRET_KEY}
+            - name: AWS_REGION
+              value: "us-west-1"
+            - name: MAX_RECOG_SERVER_THREADS
```
ustiugov, line 49:
```suggestion
            - name: MAX_SERVER_THREADS
```
and in other yml files

function-images/tests/stacking-training/manifests/service-reducer.yaml
```diff
@@ -0,0 +1,56 @@
...
+          env:
+            - name: TRANSFER_TYPE
+              value: "S3"
+            - name: AWS_ACCESS_KEY
+              value: ${AWS_ACCESS_KEY}
+            - name: AWS_SECRET_KEY
+              value: ${AWS_SECRET_KEY}
+            - name: AWS_REGION
+              value: "us-west-1"
+            - name: MAX_RECOG_SERVER_THREADS
```
ustiugov, line 49:
```suggestion
            - name: MAX_SERVER_THREADS
```

function-images/tests/stacking-training/manifests/service-metatrainer.yaml
```diff
@@ -0,0 +1,56 @@
...
+          env:
+            - name: TRANSFER_TYPE
+              value: "S3"
+            - name: AWS_ACCESS_KEY
+              value: ${AWS_ACCESS_KEY}
+            - name: AWS_SECRET_KEY
+              value: ${AWS_SECRET_KEY}
+            - name: AWS_REGION
+              value: "us-west-1"
+            - name: MAX_RECOG_SERVER_THREADS
```
ustiugov, line 49:
```suggestion
            - name: MAX_SERVER_THREADS
```

function-images/tests/stacking-training/manifests/service-driver.yaml
```diff
@@ -0,0 +1,58 @@
...
+            - name: AWS_SECRET_KEY
+              value: ${AWS_SECRET_KEY}
+            - name: AWS_REGION
+              value: "us-west-1"
+            - name: CONCURRENT_TRAINING
+              value: "false"
+            - name: TrainersNum
+              value: "4"
+            - name: ENABLE_TRACING
+              value: "true"
```
ustiugov, line 54: add `- name: MAX_SERVER_THREADS` (I guess)

ustiugov (commented):
lgtm

MBaczun:
re-generating the proto files gives me the same result. I would suggest removing the golang protos since we only use python in this benchmark, and it looks like some go library is causing problems.

ustiugov:
> re-generating the proto files gives me the same result. I would suggest removing the golang protos since we only use python in this benchmark, and it looks like some go library is causing problems.

agreed, please do. remove that generation from the makefile.

function-images/tests/stacking-training/driver/main.py
```diff
@@ -0,0 +1,351 @@
...
+                log.fatal("XDT is not yet supported")
+            return pickle.loads(response['Body'].read())
+    
+    def put_to_s3(self, pickled, key):
+        s3object = self.s3_client.Object(bucket_name=self.benchName, key=key)
+        s3object.put(Body=pickled)
+
+    def get_from_s3(self, key):
+        obj = self.s3_client.Object(bucket_name=self.benchName, key=key)
+        return obj.get()
```
ustiugov, line 227: these functions are very short, let's inline them into put() and get(). Same for other functions

function-images/tests/stacking-training/driver/main.py
```diff
@@ -0,0 +1,351 @@
...
+        log.info("Get the final outputs")
+
+        _ = self.get(outputs['model_full_key'])
+        _ = self.get(outputs['meta_predictions_key'])
+
+    # Driver code below
+    def SayHello(self, request, context):
+        log.info("Driver received a request")
+
+        self.upload_dataset()
```
ustiugov, line 313: let's replace this function with a direct call to put()

ustiugov (commented):
Few minor things to refactor:
- [x] move the manifests to `stacking-training/knative_yamls/s3/`, similarly to the video analytics benchmark.
- [x] replace upload_dataset with direct use of put(); remove get_from_s3 and put_to_s3 functions.

.github/workflows/function-composition.yml
```diff
@@ -361,3 +391,93 @@ jobs:
...
+          sleep 1m
+      - name: invoke the chain
+        run: |
+          ./bin/grpcurl -plaintext localhost:50051 helloworld.Greeter.SayHello
+      - name: show docker-compose log
+        if: ${{ always() }}
+        run: cat  ./function-images/tests/stacking-training/log_file
+
+  bench-stacking-training-knative:
+    name: Video analytics benchmark (Knative)
```
ustiugov, line 424: wrong name

.github/workflows/function-composition.yml
```diff
@@ -361,3 +391,93 @@ jobs:
...
+      - uses: actions/checkout@v2
+        with:
+          lfs: 'true'
+
+      - name: Deploy functions
+        env:
+          AWS_ACCESS_KEY: ${{ secrets.AWS_ACCESS_KEY }}
+          AWS_SECRET_KEY: ${{ secrets.AWS_SECRET_KEY }}
+        run: |
+          ./function-images/tests/kn_deploy.sh ./function-images/tests/stacking-training/manifests/service*
```
ustiugov, line 448: these functions, particularly the driver, needs some time to initialize (e.g., generate the dataset). As a result, the invoker's request times out.

Please add a sleep for 1min after the deployment

.github/workflows/function-composition.yml
```diff
@@ -361,3 +391,93 @@ jobs:
...
+      - name: Invoke functions
+        working-directory: examples/invoker
+        run: |
+          set -x
+
+          NODEPORT=$(kubectl get svc kourier-ingress -n kourier-system -o=jsonpath='{.spec.ports[0].nodePort}')
+          HOSTNAME=$(kubectl get ksvc streaming -n default -o jsonpath='{.status.url}' | cut -c8-)
+
+          echo '[ { "hostname": "$HOSTNAME" } ]' > endpoints.json
+          ./invoker -port $NODEPORT -dbg
```
ustiugov, line 467: issue with a low rate, e.g., -rps 0.3, so only 1 request is issued.

function-images/tests/stacking-training/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+RUN --mount=type=secret,id=GOPRIVATE_KEY \
+        git config --global url."https://ease-lab:$(cat /run/secrets/GOPRIVATE_KEY)@github.com/ease-lab/vhive-xdt".insteadOf "https://github.com/ease-lab/vhive-xdt"
+RUN git clone --depth 1 https://github.com/ease-lab/vhive-xdt /xdt
+
+FROM vhiveease/python-slim:latest
+COPY --from=builder /root/.local /root/.local
+COPY --from=builder /py /app
+COPY --from=builder /xdt/sdk/python/ /app
+COPY --from=builder /xdt/proto/downXDT/ /app
+COPY --from=builder /xdt/proto/upXDT/ /app
```
shyamjesal, line 44: add `COPY --from=builder /xdt/proto/crossXDT/ /app`

.github/workflows/function-composition.yml
```diff
@@ -361,3 +391,94 @@ jobs:
...
+      - name: Invoke functions
+        working-directory: examples/invoker
+        run: |
+          set -x
+
+          NODEPORT=$(kubectl get svc kourier-ingress -n kourier-system -o=jsonpath='{.spec.ports[0].nodePort}')
+          HOSTNAME=$(kubectl get ksvc streaming -n default -o jsonpath='{.status.url}' | cut -c8-)
+
+          echo '[ { "hostname": "$HOSTNAME" } ]' > endpoints.json
+          ./invoker -port $NODEPORT -dbg -rps 0.3
```
ustiugov, line 468: please add 2min sleep here as well because the benchmark takes time to run

function-images/tests/stacking-training/reducer/main.py
```diff
@@ -0,0 +1,191 @@
...
+        msg = "Driver gets key '" + key + "' from " + self.transferType
+        log.info(msg)
+        with tracing.Span(msg):
+            response = None
+            if self.transferType == S3:
+                obj = self.s3_client.Object(bucket_name=self.benchName, key=key)
+                response = obj.get()
+            elif self.transferType == XDT:
+                log.fatal("XDT is not yet supported")
+            return pickle.loads(response['Body'].read())
```
ustiugov, line 141:
```suggestion
                log.fatal("XDT is not yet supported")

            return pickle.loads(response['Body'].read())
```

function-images/tests/stacking-training/reducer/main.py
```diff
@@ -0,0 +1,191 @@
...
+        msg = "Driver uploading object with key '" + key + "' to " + self.transferType
+        log.info(msg)
+        with tracing.Span(msg):
+            pickled = pickle.dumps(obj)
+            if self.transferType == S3:
+                s3object = self.s3_client.Object(bucket_name=self.benchName, key=key)
+                s3object.put(Body=pickled)
+            elif self.transferType == XDT:
+                log.fatal("XDT is not supported")
+        return key
```
ustiugov, line 129:
```suggestion
                log.fatal("XDT is not supported")

        return key
```

ustiugov (commented):
another sleep and few more cosmetic changes (important because we are going to reuse these functions a lot)

ustiugov:
Same [problem](https://github.com/ease-lab/vhive/pull/365/checks?check_run_id=3234445264#step:8:49) with crossXDT, @MBaczun please add the same fix as for the stacking benchmark.

ustiugov:
I found an even more important [problem](https://github.com/ease-lab/vhive/pull/365/checks?check_run_id=3234445264#step:4:25): kn_deploy.sh's failure to deploy a function does not fail the test, resulting in a timeout. 

Please fix by adding `set -e` to the script.

MBaczun:
Not sure about the port being used in CI, see [here](https://github.com/ease-lab/vhive/runs/3240399207?check_suite_focus=true#step:7:21), where it uses 31080. I think this should be using port 80 every time. A few lines later when we print logs ([here](https://github.com/ease-lab/vhive/runs/3240399207?check_suite_focus=true#step:8:1)) you can see that the benchmark didn't actually do anything, it's just all of the initialisation messages. The way the invoker works it doesn't report that a weird port was used and that the connection didn't work, so the action passes.

I think the quick and easy fix is to hardcode the action to use port 80 instead of `NODEPORT=$(kubectl get svc kourier-ingress -n kourier-system -o=jsonpath='{.spec.ports[0].nodePort}')`.

.github/workflows/function-composition.yml
```diff
@@ -361,3 +395,97 @@ jobs:
...
+      - name: Invoke functions
+        working-directory: examples/invoker
+        run: |
+          set -x
+
+          NODEPORT=$(80)
+          HOSTNAME=$(kubectl get ksvc streaming -n default -o jsonpath='{.status.url}' | cut -c8-)
+
+          echo '[ { "hostname": "$HOSTNAME" } ]' > endpoints.json
+          ./invoker -port $NODEPORT -dbg -rps 0.3
```
ustiugov, line 474: the invoker manages to invoke it 2x, please adjust its runtime so it waits for 2 min (remove the sleep though) and `-rps` so that it only makes a single invocation

ustiugov (commented):
revert the nodeport, configure the invoker

.github/workflows/function-composition.yml
```diff
@@ -361,3 +395,97 @@ jobs:
...
+        working-directory: examples/invoker
+        run: make invoker
+
+      - name: Invoke functions
+        working-directory: examples/invoker
+        run: |
+          set -x
+
+          NODEPORT=$(kubectl get svc kourier-ingress -n kourier-system -o=jsonpath='{.spec.ports[0].nodePort}')
+          HOSTNAME=$(kubectl get ksvc streaming -n default -o jsonpath='{.status.url}' | cut -c8-)
```
ustiugov, line 471:
```suggestion
          HOSTNAME=$(kubectl get ksvc driver -n default -o jsonpath='{.status.url}' | cut -c8-)
```

PR #370, opened 2021-08-02
gg port

cbfariasc:
To test using stock Knative:
1) Run the stock-only setup scripts (https://github.com/faromero/vhive/blob/main/docs/developers_guide.md#testing-stock-knative-setup-or-images)
2) Run "kn service apply gg-port-0 --filename configs/knative_workloads/gg_port.yaml"
Note: This might require for you to run sudo chown $(id -u):$(id -g) $HOME/.kube/config due to file access permissions if the config file can't be loaded
3) Run "curl -v -X POST url" using the url outputted by the above command (It should respond with a JSON error)

To test using gg input (in progress):
1) Clone vhive_port branch from http://github.com/faromero/gg
2) Run setup (https://github.com/StanfordSNR/gg#build-directions)
3) Enter a test directory, such as the one in tests/hello_world, or mosh directory and run gg init, and gg infer make
4) Run "gg force --jobs 1 --engine vhive=(IP or url):port executable" to compile the executable
Note: This is where I have been trying to get a working url or IP for the ingress gateway. I have been testing with curl commands as I tried to debug, so that would be another way to test for a working connection with the ingress gateway.

### DIFF
--- a/.github/workflows/function-composition.yml
+++ b/.github/workflows/function-composition.yml
@@ -171,6 +171,7 @@ jobs:
           ./bin/grpcurl -max-time 10 -plaintext localhost:3031 helloworld.Greeter.SayHello
 
       - name: show docker-compose log
+        if: ${{ always() }}
         run: cat function-images/tests/chained-function-serving/log_file
 
   bench-serving:
@@ -327,14 +328,25 @@ jobs:
         run: |
           ./function-images/tests/kn_deploy.sh ./function-images/tests/video-analytics/knative_yamls/${{ matrix.transfer-type }}/service*
 
+      - uses: actions/setup-go@v2
+        with:
+          go-version: 1.16
+
+      - name: Build invoker
+        working-directory: examples/invoker
+        run: make invoker
+
       - name: Invoke functions
+        working-directory: examples/invoker
         run: |
           set -x
 
           NODEPORT=$(kubectl get svc kourier-ingress -n kourier-system -o=jsonpath='{.spec.ports[0].nodePort}')
           HOSTNAME=$(kubectl get ksvc streaming -n default -o jsonpath='{.status.url}' | cut -c8-)
 
-          ./bin/grpcurl -plaintext $HOSTNAME:$NODEPORT helloworld.Greeter.SayHello
+          echo '[ { "hostname": "$HOSTNAME" } ]' > endpoints.json
+          ./invoker -port $NODEPORT -dbg
+          cat rps*lat.csv
 
       - name: Print logs
         if: ${{ always() }}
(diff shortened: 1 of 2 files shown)


function-images/gg_port/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+COPY common /common
+COPY --from=builder_workload /root/.local /root/.local
+RUN apk add libstdc++ --update --no-cache
+ENV APP_HOME /app
+WORKDIR $APP_HOME
+EXPOSE 50051
+
+STOPSIGNAL SIGKILL
+
+#CMD exec gunicorn --bind :50051 --workers 1 app:app
```
ustiugov, line 17: please remove the comment

function-images/gg_port/greeter_client.py
```diff
@@ -0,0 +1,45 @@
...
+import helloworld_pb2_grpc
+import json
+import sys
+from minio import Minio
+
+def run():
+    with grpc.insecure_channel('localhost:50051') as channel:
+        stub = helloworld_pb2_grpc.GreeterStub(channel)
+
+        minioClient = Minio("10.138.0.27:9000",
```
ustiugov, line 16: if possible, would be really good to add AWS S3 as an alternative to MinIO.

function-images/gg_port/requirements.txt
```diff
@@ -0,0 +1,4 @@
+futures
```
ustiugov, line 1: please fix the versions to the current ones to avoid python incompatibilities

configs/knative_workloads/gg_port.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: serving.knative.dev/v1
+kind: Service
+metadata:
+  name: gg-port-0
```
ustiugov, line 4: I suggest choosing a better name, maybe `gg-worker`?

ustiugov (changes requested):
great job, folks! Thank you for the contribution!

Before merging, please do the following cleaning and cosmetics:
- [ ] revert .gitignore, docker_build
- [x] remove all binary files (*.pyc, *tar.gz, *.zip, invoker binary, runc, ...). If necessary, use Git LFS (see an example [here](https://github.com/ease-lab/vhive/tree/main/bin)). For example, for `gg-execute-static` if you do not submit its sources. However, it would be better to re-build it, maybe from another repo. 
- [ ] remove changes in image-rotate and minio-scripts. I guess they are not related to the PR
- [ ] please add MIT license to all source files (dockerfile, bash, python, etc.)
- [ ] add a README describing how to build and run any simple example with gg in the PR's description

After applying all of these changes, please squash all commits into a single commit, like described [here](https://github.com/ease-lab/vhive/blob/main/docs/contributing_to_vhive.md#squashing-commits). 

Once the revision is done, we can come up with a plan for integrating gg into our CI. We can help with that.

PR #371, opened 2021-08-03
Adds hyperparameter tuning halving benchmark

MBaczun:
Hyperparameter Tuning Halving is a benchmark comprised of two functions: the Driver and the Trainer. Models with varying hyperparameters are trained repeatedly over a subset of the dataset, and only the top performing 50% of models make it to the next iteration. This is repeated until only one model remains, and with each iteration the sample rate over the dataset is increased such that larger samples are utilised as the benchmark progresses. 

The Driver manages the core loop and maintains all of the models being trained. It issues calls to the Trainer function to train individual models, and then decides which models make it to the next iteration given their score as reported by the trainer.

The Trainer trains a single model by fitting to a subset of the dataset (the sample rate being provided by the Driver). The score of the model is returned to the Driver. Note that multiple instances of the Trainer could be active simultaneously as each instance independently trains a separate model.

Signed-off-by: Michal Baczun <s1743497@ed.ac.uk>


function-images/tests/tuning-halving/driver/main.py
```diff
@@ -0,0 +1,262 @@
...
+        msg = "Driver uploading object with key '" + key + "' to " + self.transferType
+        log.info(msg)
+        with tracing.Span(msg):
+            pickled = pickle.dumps(obj)
+            if self.transferType == S3:
+                s3object = self.s3_client.Object(bucket_name=self.benchName, key=key)
+                s3object.put(Body=pickled)
+            elif self.transferType == XDT:
+                log.fatal("XDT is not supported")
+        return key
```
ustiugov, line 140: please apply the same comments as in the previour PR to get() and put()

function-images/tests/tuning-halving/driver/main.py
```diff
@@ -0,0 +1,262 @@
...
+                    self.train({
+                        'dataset_key': event['dataset_key'],
+                        'model_config': model_config,
+                        'count': count,
+                        'sample_rate': sample_rate
+                    })
+                )
+
+            # Keep models with the best score
+            top_number = len(training_responses)//2
```
ustiugov, line 221: is `//` a typo?

function-images/tests/tuning-halving/manifests/service-driver.yaml
```diff
@@ -0,0 +1,60 @@
...
+            - name: AWS_ACCESS_KEY
+              value: ${AWS_ACCESS_KEY}
+            - name: AWS_SECRET_KEY
+              value: ${AWS_SECRET_KEY}
+            - name: AWS_REGION
+              value: "us-west-1"
+            - name: CONCURRENT_TRAINING
+              value: "false"
+            - name: TrainersNum
+              value: "4"
```
ustiugov, line 52: this is unused in this bench, right?

function-images/tests/tuning-halving/trainer/main.py
```diff
@@ -0,0 +1,211 @@
...
+        msg = "Driver gets key '" + key + "' from " + self.transferType
+        log.info(msg)
+        with tracing.Span(msg):
+            response = None
+            if self.transferType == S3:
+                obj = self.s3_client.Object(bucket_name=self.benchName, key=key)
+                response = obj.get()
+            elif self.transferType == XDT:
+                log.fatal("XDT is not yet supported")
+            return pickle.loads(response['Body'].read())
```
ustiugov, line 147: same comment here

ustiugov (changes requested):
lgtm

- [ ] verify that the output is exactly the same as for the original script.
- [ ] add CI
- [ ] format put() and get() as in the other PR

function-images/tests/tuning-halving/driver/main.py
MBaczun: I don't think so. I didn't type this, it's from the original code, I thought it's just integer division

MBaczun:
looks like there's an issue with the invoker where it doesn't wait for the driver to finish

ustiugov:
the invoker terminates without waiting by design. just increase its exec time and decrease the rps so that it only issues a single request in the whole run

function-images/tests/tuning-halving/driver/main.py
```diff
@@ -0,0 +1,264 @@
...
+        }
+        models_config = {
+            'models': [
+                {
+                    'model': 'RandomForestRegressor',
+                    'params': hyperparam
+                } for hyperparam in generate_hyperparam_sets(hyperparam_config['params'])
+            ]
+        }
+        self.put(dataset, 'dataset_key')
```
ustiugov, line 174: BUG: need to record the key that `put()` returns and pass it. make sure that every function does it. otherwise XDT would not work

function-images/tests/tuning-halving/driver/main.py
```diff
@@ -0,0 +1,264 @@
...
+        return {
+            'dataset_key': 'dataset_key',
```
ustiugov, line 176: pass the key here instead of the constant

function-images/tests/tuning-halving/trainer/main.py
```diff
@@ -0,0 +1,213 @@
...
+                X_sampled, y_sampled = X[sampled_index], y[sampled_index]
+            y_pred = cross_val_predict(model, X_sampled, y_sampled, cv=5)
+            model.fit(X_sampled, y_sampled)
+            score = roc_auc_score(y_sampled, y_pred)
+            log.info(f"{model_config['model']}, params: {model_config['params']}, dataset size: {len(y_sampled)},score: {score}")
+
+        # Write to S3
+        model_key = f"model_{count}"
+        pred_key = f"pred_model_{count}"
+        self.put(model, model_key)
```
ustiugov, line 181: same bug here

ustiugov (changes requested):
need a minor bug fix and good to go

ustiugov:
also, need a brief description of this benchmark in the PR's descr.

ustiugov:
the code moved to a separate [repo](https://github.com/ease-lab/vSwarm).

PR #374, opened 2021-08-04
Add mare test word-count

boramalper:
Signed-off-by: Bora M. Alper <bora@boramalper.org>


ustiugov:
closed in favor of Corral, which is already merged.

PR #377, opened 2021-08-05
Fixed bug in stacking benchmark

ustiugov:
Dataset key must be taken from the get() return value.


MBaczun (approved):
lgtm

PR #378, opened 2021-08-05
Storage module

MBaczun:
Includes:
- Golang module
   - Put and PutFile methods (PutFile is the same as Put but allows user to upload a file by providing the file object instead of a byte array)
   -  Get
   - Golang Tests for Put and Get on s3 and elasticache (Elasticache test only works if you are on an aws host)
- Python module
   - Put and Get functions for s3 and elasticache. Includes optional `doPickle` field which lets the user disable pickling, this is sometimes necessary since only python has pickling so if one `put`s a pickled file with python and `get`s it with golang then they would not be able to un-pickle it.
   - Python tests for Put and Get with s3 and elasticache (again, elasticache only works on ec2)
- Updated `stacking-training` and `video-analytics` benchmarks to use the module.

Note that the module tests are not included in the CI since they only pass if run on EC2 machines.
 
Signed-off-by: Michal Baczun <s1743497@ed.ac.uk>


utils/storage/go/storage.go
```diff
@@ -0,0 +1,178 @@
...
+			Body:   file,
+		})
+		log.Infof("S3 upload complete")
+		if err != nil {
+			log.Fatalf("Failed to upload bytes to s3: %s", err)
+		}
+		return key
+
+	} else if transferType == ELASTICACHE {
+		log.Fatalf("File transfer via ElastiCahce currently unsupported, please use []bytes: `Put(key string, payloadData []byte) string`")
```
ustiugov, line 145:
```suggestion
		log.Fatalf("File transfer via ElastiCache is currently not supported, please use []bytes: `Put(key string, payloadData []byte) string`")
```

utils/storage/python/requirements.txt
```diff
@@ -0,0 +1,2 @@
+boto3==1.18.1
+redis
```
ustiugov, line 2: fix the version

utils/storage/go/storage.go
```diff
@@ -0,0 +1,178 @@
...
+			panic(err)
+		}
+		return key
+	} else {
+		log.Fatalf("Unsupported transfer type: %s", transferType)
+	}
+	return ""
+}
+
+// Put uploads the payload to the storage service using the provided key
```
ustiugov, line 128: do we actually need this method? there is none like that in the python module

utils/storage/python/storage.py
```diff
@@ -0,0 +1,104 @@
...
+            service_name='s3',
+            region_name=os.getenv("AWS_REGION", 'us-west-1'),
+            aws_access_key_id=AWS_ID,
+            aws_secret_access_key=AWS_SECRET
+        )
+    elif transferType == ELASTICACHE:
+        elasticache_client = redis.Redis.from_url(bucket)
+
+# `put` uploads the payload to the storage service using the provided key
+def put(key, obj, dontPickle = False):
```
ustiugov, line 64: I suggest using a non-neg argument to avoid confusion: `toPickle` default True

ustiugov (changes requested):
- [x] remove the *2 suffix
- [x] fix the redis dep version
- [x] use toPickle instead of dontPickle
- [ ] add S3 unit tests to the CI (no need to test ElastiCache as of now)

utils/storage/go/storage.go
```diff
@@ -0,0 +1,178 @@
...
+	benchName = bucket
+	awsAccessKey, ok := os.LookupEnv("AWS_ACCESS_KEY")
+	if ok {
+		AKID = awsAccessKey
+	}
+	awsSecretKey, ok := os.LookupEnv("AWS_SECRET_KEY")
+	if ok {
+		SECRET_KEY = awsSecretKey
+	}
+	AWS_S3_REGION = "us-west-1"
```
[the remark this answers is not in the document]
MBaczun: This happens on the line bellow. us-west-1 is just used as a default value

ustiugov (approved):
lgtm; restarting the runners now

PR #396, opened 2021-09-02
Scripts to start gvisor-cri-runners inside multipass-VMs

nathanieltornow:
- Modified `scripts/github_runner/setup_runner_host.sh` to download multipass-packages when `multipass` flag is set.
- Modified `scripts/gihub_runner/start_runners.sh` to start gv-runners inside multipass VMs when `gvisor-cri` is set as runner_group. The script now depends on the new script `scripts/gihub_runner/start_bare-metal_runner.sh`.
- Added script `scripts/gihub_runner/start_bare-metal_runner.sh`, which starts a runner on the machine where the script is being run. This script follows the standard steps to start a GH-runner and has no dependencies on other scripts.

### DIFF
--- a/scripts/github_runner/start_bare-metal_runner.sh
+++ b/scripts/github_runner/start_bare-metal_runner.sh
@@ -2,7 +2,7 @@
 
 # MIT License
 #
-# Copyright (c) 2020 Dmitrii Ustiugov, Shyam Jesalpura and EASE lab
+# Copyright (c) 2020 Nathaniel Tornow and EASE lab
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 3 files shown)


scripts/github_runner/setup_runner_host.sh
```diff
@@ -33,6 +36,13 @@ sudo apt-get install -y \
     jq \
     software-properties-common >> /dev/null
 
+if [ "${MULTIPASS}" == "multipass" ]; then
+	sudo apt-get -y install snapd >> /dev/null
+	sleep 10s
+	sudo snap install multipass >> /dev/null
+	exit 0
```
ustiugov, line 43: why sleep and why exit 0? because simplicity in our bash scripts is preferable

scripts/github_runner/start_bare-metal_runner.sh
```diff
@@ -0,0 +1,46 @@
...
+
+cd
+git clone ${URL}
+cd vhive
+./scripts/cloudlab/setup_node.sh ${SANDBOX}
+
+cd
+mkdir actions-runner && cd actions-runner
+
+curl -o actions-runner-linux-x64-2.280.3.tar.gz -L https://github.com/actions/runner/releases/download/v2.280.3/actions-runner-linux-x64-2.280.3.tar.gz
```
ustiugov, line 40: please make it fetch the latest version, look at how it is done inside the docker runner

ustiugov (changes requested):
lgtm but need to apply some minor changes

scripts/github_runner/setup_runner_host.sh
nathanieltornow: Exit 0 because nothing else has to be installed, so just to go out of the script (thought it would be more readable than a if-else). We can of course also go for a version where everything is installed without choice between multipass or docker.

sleep 10s because I had some issues with snap taking some time to start up and be ready for installation.

scripts/github_runner/start_bare-metal_runner.sh
shyamjesal: checkout [this](https://github.com/ease-lab/vhive/blob/2a0979407942294c3073e0c03008b0936bd202d2/scripts/github_runner/setup_runner.sh#L47).

scripts/github_runner/setup_runner_host.sh
ustiugov: Ok. Yes let's install everything without a choice

scripts/github_runner/start_bare-metal_runner.sh
nathanieltornow: Fixed

scripts/github_runner/setup_runner_host.sh
nathanieltornow: Since it just happens sporadically, I also removed the sleep, but one can see if it fails, and just restart the script.

ustiugov: I think it's better to leave the sleep if it's more stable this way

ustiugov: so what's about the sleep? restarting is not a good idea because self-hosted runners are already not quite stable

nathanieltornow: Yes, it is now back in.

ustiugov (approved):

Issue #402, opened 2021-09-24
Quick start script dependency missing


cvetkovic:
**Describe the bug**
While running the II.1.4 command described in the vHive Quickstart, an issue as shown in the picture happens (a broken URL). Further steps cannot be executed because 'cri-tools_1.17.0~3_amd64.deb' are not installed. It worked up until today.
![Screenshot from 2021-09-24 11-55-24](https://user-images.githubusercontent.com/16463659/134677638-5fba653e-c4ca-4312-acff-03a7ae3c471b.png)

cvetkovic:
This issue is caused by apt-get install cri-tools command in the 'scripts/install_stock.sh' script. This command worked from Switzerland, but not from the CloudLab server in Utah, because the HTTP request got mirrored to a repository at Purdue. The apt-get directory was inconsistent at the time this issue was created.

ustiugov:
this issue needs investigation and probably a proper fix

ustiugov:
The issue is related to the fact that skopeo (the tool Amory introduced for using local registry) dropped support for Ubuntu 18 packages. 

A proper solution is to migrate to Ubuntu 20. In the mean time, a hotfix will be provided.

ustiugov:
hotfix is to be introduced by #415

PR #411, opened 2021-09-30
Fixed invoker & deployer in the quickstart guide

ustiugov:
The invoker and the deployer tools need to be built from their enclosing folders. This bug was introduced by our go.mod files optimizations. 

### DIFF
--- a/docs/quickstart_guide.md
+++ b/docs/quickstart_guide.md
@@ -241,7 +241,7 @@ This section is only for synchronous (i.e., Knative Serving) functions. Please r
 1. Optionally, configure the types and the number of functions to deploy in `examples/deployer/functions.json`.
 2. Run the deployer client:
     ```bash
-    source /etc/profile && go run examples/deployer/client.go
+    source /etc/profile && pushd ./examples/deployer && go build && popd && ./examples/deployer/deployer
     ```
     > **BEWARE:**
     >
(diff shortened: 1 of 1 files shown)


ustiugov:
@cvetkovic could you please review and verify if this docs fix is sufficient?

It should close #402.

cvetkovic (changes requested):
Invoker now works. I have added a note on the endpoints.json file in case one finds it to be missing.

docs/quickstart_guide.md
```diff
@@ -258,11 +258,13 @@ This section is only for synchronous (i.e., Knative Serving) functions. Please r
 
 1. Run the invoker client:
     ```bash
-    go run examples/invoker/client.go
+    pushd ./examples/invoker && go build && popd && ./examples/invoker/invoker
     ```
 
     > **Note:**
     >
+    > The aforementioned command can only be run on the same node where the deployer command was invoked. In order to run the invoker on other nodes, the generated `endpoints.json` file has to be copied manually to the root directory of vHive.
```
ustiugov, line 266: @cvetkovic can you instead say: copy the file produced by the deployer and run the invoker with the XX argument to specify the path and the name of that file

ustiugov:
@cvetkovic can you please also fix the commit linter error? You need to sign your commits (common practice in oss repos) by running 'git commit -s --amend' then force-push your commit

docs/quickstart_guide.md
cvetkovic: Done. No need to specify any arguments if the JSON file is copied to vHive root.

PR #432, opened 2021-11-11
Unlock memory manager upon returning error

amohoste:
Currently, the memory manager is not always unlocked when an error is returned.


ustiugov (changes requested):
Please fix the failing link (see the Linters Check). The correct one is https://github.com/cloudevents/spec/blob/v1.0.1/primer.md#cloudevents-concepts

Other than that, approved

ustiugov (approved):

PR #433, opened 2021-11-17
Replace iptables by nftables for microVM networking

amohoste:
## Description
This commit moves the firecracker microVM network configuration from iptables to [nftables](https://wiki.nftables.org/wiki-nftables/index.php/Main_Page) which is configured directly through a [third party go package](https://github.com/google/nftables). The Github runner may need to be updated to add the additional dependency in order for the tests to run properly. This PR is the first of several PRs aiming to improve snapshotting functionality in vHive.


ustiugov (approved):
lgtm

PR #439, opened 2021-11-23
Add image manager to manage container image pulling

amohoste:
This PR adds an image manager which manages and serializes pulling of identical container images. This avoids unnecessary network congestion as well as a starting point to managing which container images are kept cached on a node.

### DIFF
--- a/ctrimages/imageManager.go
+++ b/ctrimages/imageManager.go
@@ -1,0 +1,143 @@
+// Package ctrimages provides an image manager that manages and caches container images.
+package ctrimages
+
+import (
+	"context"
+	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/remotes/docker"
+	log "github.com/sirupsen/logrus"
+	"net"
+	"net/http"
+	"net/url"
+	"strings"
+	"sync"
+)
+
+// ImageState is used to synchronize image pulling to avoid pulling the same image multiple times concurrently.
+type ImageState struct {
+	sync.Mutex
+	pulled bool
+}
+
+// NewImageState creates a new ImageState object that can be used to synchronize image pulling.
+func NewImageState() *ImageState {
+	state := new(ImageState)
+	state.pulled = false
+	return state
+}
+
+// ImageManager manages the images that have been pulled to the node.
+type ImageManager struct {
+	sync.Mutex
+	snapshotter  string						 // image snapshotter
+	cachedImages map[string]containerd.Image // Cached container images
+	imageStates  map[string]*ImageState
+	client       *containerd.Client
+}
+
+// NewImageManager creates a new imagemanager that can be used to fetch container images.
+func NewImageManager(client *containerd.Client, snapshotter string) *ImageManager {
+	log.Info("Creating image manager")
+	manager := new(ImageManager)
+	manager.snapshotter = snapshotter
+	manager.cachedImages = make(map[string]containerd.Image)
+	manager.imageStates = make(map[string]*ImageState)
+	manager.client = client
+	return manager
+}
+
+// pullImage fetches an image and adds it to the cached image list
+func (mgr *ImageManager) pullImage(ctx context.Context, imageName string) error {
+	var err error
+	var image containerd.Image
+
+	imageURL := getImageURL(imageName)
+	local, _ := isLocalDomain(imageURL)
+	if local {
+		// Pull local image using HTTP
+		resolver := docker.NewResolver(docker.ResolverOptions{
+			Client: http.DefaultClient,
+			Hosts: docker.ConfigureDefaultRegistries(
+				docker.WithPlainHTTP(docker.MatchAllHosts),
+			),
+		})
+		image, err = mgr.client.Pull(ctx, imageURL,
+			containerd.WithPullUnpack,
+			containerd.WithPullSnapshotter(mgr.snapshotter),
+			containerd.WithResolver(resolver),
+		)
+	} else {
+		// Pull remote image
+		image, err = mgr.client.Pull(ctx, imageURL,
+			containerd.WithPullUnpack,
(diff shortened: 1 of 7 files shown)


scripts/install_pmutools.sh
```diff
@@ -32,4 +32,4 @@ sudo git clone https://github.com/andikleen/pmu-tools /usr/local/pmu-tools
 sudo sysctl -w kernel.perf_event_paranoid=-1
 
 # first run, download essential files
-/usr/local/pmu-tools/toplev --print > /dev/null
+#/usr/local/pmu-tools/toplev --print > /dev/null
```
ustiugov, line 35: this works, please revert

ctrimages/imageManager.go
```diff
@@ -0,0 +1,143 @@
...
+func NewImageState() *ImageState {
+	state := new(ImageState)
+	state.pulled = false
+	return state
+}
+
+// ImageManager manages the images that have been pulled to the node.
+type ImageManager struct {
+	sync.Mutex
+	snapshotter  string						 // image snapshotter
```
ustiugov, line 32: please fix the indent in the comment

ctrimages/imageManager.go
```diff
@@ -0,0 +1,143 @@
...
+	cachedImages map[string]containerd.Image // Cached container images
+	imageStates  map[string]*ImageState
+	client       *containerd.Client
+}
+
+// NewImageManager creates a new imagemanager that can be used to fetch container images.
```
ustiugov, line 38:
```suggestion
// NewImageManager creates a new image manager that can be used to fetch container images.
```

ctrimages/imageManager.go
```diff
@@ -0,0 +1,143 @@
+// Package ctrimages provides an image manager that manages and caches container images.
```
ustiugov, line 1: add a copyright please

ctrimages/imageManager.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/containerd/containerd/remotes/docker"
+	log "github.com/sirupsen/logrus"
+	"net"
+	"net/http"
+	"net/url"
+	"strings"
+	"sync"
+)
+
+// ImageState is used to synchronize image pulling to avoid pulling the same image multiple times concurrently.
```
ustiugov, line 16:
```suggestion
// ImageState is used for synchronization to avoid pulling the same image multiple times concurrently.
```

ctrimages/imageManager.go
```diff
@@ -0,0 +1,143 @@
...
+type ImageState struct {
+	sync.Mutex
+	pulled bool
```
ustiugov, line 19:
```suggestion
	isCached bool
```

ctrimages/imageManager.go
```diff
@@ -0,0 +1,143 @@
...
+		// Pull remote image
+		image, err = mgr.client.Pull(ctx, imageURL,
+			containerd.WithPullUnpack,
+			containerd.WithPullSnapshotter(mgr.snapshotter),
+		)
+	}
+	if err != nil {
+		return err
+	}
+	mgr.Lock()
```
ustiugov, line 79: that's a bug. you need to protect the whole function body with a lock.

ustiugov, line 79: that's a bug. you need to protect the whole function body with a lock.

ctrimages/imageManager.go
```diff
@@ -0,0 +1,143 @@
...
+
+// GetImage fetches an image that can be used to create a container using containerd
+func (mgr *ImageManager) GetImage(ctx context.Context, imageName string) (*containerd.Image, error) {
+	mgr.Lock()
+	imgState, found := mgr.imageStates[imageName]
+	if !found {
+		imgState = NewImageState()
+		mgr.imageStates[imageName] = imgState
+	}
+	mgr.Unlock()
```
ustiugov, line 93: style: use `defer ...` right after obtaining the lock

ctrimages/imageManager.go
```diff
@@ -0,0 +1,143 @@
...
+
+	// Pull image if necessary
+	imgState.Lock()
```
ustiugov, line 96: this lock is redundant, there is a lock inside that function already

ctrimages/imageManager.go
```diff
@@ -0,0 +1,143 @@
...
+		return err
+	}
+	mgr.Lock()
+	mgr.cachedImages[imageName] = image
+	mgr.Unlock()
+	return nil
+}
+
+// GetImage fetches an image that can be used to create a container using containerd
+func (mgr *ImageManager) GetImage(ctx context.Context, imageName string) (*containerd.Image, error) {
```
ustiugov, line 86: I have a feeling that only this function needs a lock and mgr.Pull() should not have a lock. is it right?

ustiugov (changes requested):
High-level it's good, thank you. However, I am concerned that there are concurrency bugs. Need to do the following:

- [ ] create a separate folder under ctriface/image/manager. No need to create a go.mod file though
- [ ] add unit tests that would model concurrent image pull situation (please use an image hosted on gcr.io not in Docker because there the number of pulls is not limited)
- [ ] add copyrights
- [ ] minor refactoring

ctrimages/imageManager.go
amohoste: Not sure if I see the bug, the imagemanager mutex is only used to synchronise access to the imagemanager internal maps. There then also is an additional lock for each image to synchronize the pulling of that specific image. Although I agree a test would be good to have.

ustiugov: please go over all the comments, I think the concurrency mechanism needs to be re-visited for more simplicity and testing, of course

amohoste:
I think we can close this one since all changes along with tests will be incorporated through #465

PR #532, opened 2022-05-11
Fix k8s cluster boot inside cri test runner

shyamjesal:
## Summary

Fixed the kind runner bootup.

added `configs/kind/kubeadm.conf` to boot kind cluster without cgroup errors.
fixed `scripts/cluster/create_one_node_cluster.sh` to use the above config while executing `kubeadm init`


ustiugov:
also, please squash the commits

PR #573, opened 2022-07-18
Cluster configs for large-scale experiments

Lightxyz:
## Summary

In this PR, we aim to fine-tune Knative configs to enable large-scale cluster experiments in vHive.

TBC

### DIFF
--- a/scripts/install_stock.sh
+++ b/scripts/install_stock.sh
@@ -51,6 +51,19 @@ sudo sh -c "echo 'deb http://apt.kubernetes.io/ kubernetes-xenial main' > /etc/a
 sudo apt-get update >> /dev/null
 sudo apt-get -y install cri-tools ebtables ethtool kubeadm=$K8S_VERSION kubectl=$K8S_VERSION kubelet=$K8S_VERSION kubernetes-cni >> /dev/null
 
+# Copy CNI configs
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+ROOT="$( cd $DIR && cd .. && pwd)"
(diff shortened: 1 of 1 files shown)


scripts/cluster/create_multinode_cluster.sh
```diff
@@ -32,7 +34,9 @@ Environment="KUBELET_EXTRA_ARGS=--container-runtime=remote --runtime-request-tim
 EOF'
 sudo systemctl daemon-reload
 
-sudo kubeadm init --ignore-preflight-errors=all --cri-socket /run/containerd/containerd.sock --pod-network-cidr=192.168.0.0/16
+sudo kubeadm init --v=7 --apiserver-advertise-address=10.0.1.1 --cri-socket /run/containerd/containerd.sock --pod-network-cidr=10.168.0.0/16
```
ustiugov, line 37: @HongyuHe @Lightxyz what these arguments are and what was the reason for the change?

scripts/turbo_boost.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
```
ustiugov, line 1: This should be part of loader IMO.

ustiugov (commented):
Going to rebase this PR soon.

scripts/cluster/kubeadm.yaml
```diff
@@ -0,0 +1,55 @@
+apiVersion: kubeadm.k8s.io/v1beta2
```
ustiugov: @HongyuHe I realize you copied this yaml from your local setup. Can you please explain how it is different from the default one?

ustiugov: ok, this file is deleted in the next commit

PR #594, opened 2022-09-02
Re-structure Knative and Istio YAML manifests usage and storage

ustiugov:
## Summary

Previously, we fetched the Knative & net-istio YAML files for container-based  deployments from Internet because we never had to change them. Instead, we decided to store these files in the repo (and tweak them if necessary, using the `yq` tool). Hence, I moved the vHive-specific and stock YAML files to `configs/knative_yamls/{vhive, stock}/` folders, correspondingly. 

Other changes in this PR:
* Changed the pod network IP address (CIDR) range, as the previous range overlapped with other services in CloudLab. Authored by @HongyuHe 
* Untaint the master node to allow "control plane" services (namely, Istio, Knative activator, autoscaler, and others) to be scheduled on the master node. The reason for this is that the master node in our experiments is underutilized and moving these services there should increase measurement accuracy. We may revisit this decision in future, e.g., in case the master node faces overload.
* Added a script to disable turbo boost. 

## Warning :warning:

We noticed that pod network IP range might need to be chosen differently for different cluster infrastructure. As we mostly use CloudLab for these experiments, we fine-tuned the range to its configuration. We encourage the vHive users to report any problems they face related to the IP range selection to make vHive setup more general.


ustiugov:
@HongyuHe could you please review the auto-generated YAML files? The generation script will be moved to the loader repo so it is there only for reference.

HongyuHe:
> @HongyuHe could you please review the auto-generated YAML files? The generation script will be moved to the loader repo so it is there only for reference.

Thank you very much indeed! 

But as I said, I can't tell whether it works or not by just eyeballing, since I don't remember everything by heart. 

The original setup in loader has been used in practice a zillion times, and it works. So please examining it by setting up clusters e2e and run experiments with the loader.

Thanks again!

ustiugov:
Thank you. I will test everything, no worries

scripts/utils/turbo_boost.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
```
cvetkovic, line 1: The path to this script seems wrong. In the loader repo, the address is `scripts/turbo_boost.sh`. What should we do in the end?

cvetkovic (changes requested):

scripts/utils/turbo_boost.sh
cvetkovic: Ignore this comment. I have just seen that you changed the path on the new PR to my PR.

cvetkovic (dismissed):

leokondrashov:
Replaced by #888.

PR #603, opened 2022-10-04
Fix one node install taint issue

Joffref:
It seems that the following taint is applied to the single node `node-role.kubernetes.io/control-plane`. It causes deployment suspension due to pod affinity.
### Image for reference
`kubectl get nodes`
image
`kubectl describe node vhive-sched-bench | grep Taints`
image

### DIFF
--- a/scripts/cluster/create_one_node_cluster.sh
+++ b/scripts/cluster/create_one_node_cluster.sh
@@ -61,5 +61,6 @@ fi
 
 # Untaint master (allow pods to be scheduled on master) 
 kubectl taint nodes --all node-role.kubernetes.io/master-
+kubectl taint nodes --all node-role.kubernetes.io/control-plane-
 
 $DIR/setup_master_node.sh $STOCK_CONTAINERD


Joffref:
I've follow the getting started documentation on how to deploy vHive on a single node and I clearly don't know why this label is assigned to the node.

ustiugov:
@Joffref is this problem still relevant? if so, please provide the exact steps how we can reproduce it.

Joffref:
This issue clearly depends on #606. Indeed, [since few versions master has been renamed to control-plane in kubernetes](https://github.com/kubernetes/kubeadm/issues/2200) thus as said in #606 kubeadm download is not handled by the install scripted forcing the user to manually install kubeadm. Causing this issue as the version is not the correct one.
I'm closing this pull request since this just a result of a root cause.

ustiugov:
#606 is a sporadic bug, hence it cannot be the root cause of your problem. otherwise please elaborate. 

@cvetkovic can you help make sense out of this PR/issue? I am still missing something here

Joffref:
During the installation, kubeam failed to install as described in the issue I mentioned. So, I've downloaded kubeadm by my own, causing a version gap thus a taint problem.

I've infer that kubeam has to be installed before running the getting started, which is not the case 😅

cvetkovic:
> #606 is a sporadic bug, hence it cannot be the root cause of your problem. otherwise please elaborate.
> 
> @cvetkovic can you help make sense out of this PR/issue? I am still missing something here

I suspect the mirror from which the `kubeadm` installation is pulled does not contain the version we used. This is something I have experienced in practice on Cloudlab. One geolocation has it as it uses a different mirror, the other does not.

@Joffref Can you manually check if the package manager can fetch `kubeadm` and `kubectl`, as in the `https://github.com/vhive-serverless/vHive/blob/main/scripts/install_stock.sh#L52`.

Also, this PR should not be merged, as we still do not support Kubernetes v1.25. We currently use v1.23.

Joffref:
@cvetkovic Unfortunately, I'm no more using this OS thus I can't test.

ustiugov:
I am closing this issue. Please re-open if it becomes relevant.

PR #624, opened 2022-10-13
When pulling images check for stale cached images

biecho:
## Summary

When vHive does pull images, it must check in its local cache if an image is stale. vHive should do this without pulling the whole image again. `skopeo` is used to only inspect the metadata from a remote image.

### DIFF
--- a/ctriface/iface.go
+++ b/ctriface/iface.go
@@ -319,40 +319,68 @@ func getImageURL(image string) string {
 
 func (o *Orchestrator) getImage(ctx context.Context, imageName string) (*containerd.Image, error) {
 	image, found := o.cachedImages[imageName]
-	if !found {
-		var err error
-		log.Debug(fmt.Sprintf("Pulling image %s", imageName))
 
-		imageURL := getImageURL(imageName)
-		local, _ := isLocalDomain(imageURL)
-		if local {
-			// Pull local image using HTTP
-			resolver := docker.NewResolver(docker.ResolverOptions{
-				Client: http.DefaultClient,
-				Hosts: docker.ConfigureDefaultRegistries(
-					docker.WithPlainHTTP(docker.MatchAllHosts),
-				),
-			})
-			image, err = o.client.Pull(ctx, imageURL,
-				containerd.WithPullUnpack,
-				containerd.WithPullSnapshotter(o.snapshotter),
-				containerd.WithResolver(resolver),
-			)
-		} else {
-			// Pull remote image
-			image, err = o.client.Pull(ctx, imageURL,
-				containerd.WithPullUnpack,
-				containerd.WithPullSnapshotter(o.snapshotter),
-			)
-		}
+	imageUrl := getImageURL(imageName)
+	if !found || imageIsOutdated(image, imageUrl) {
+		image, err := o.pullImage(ctx, imageUrl)
 
 		if err != nil {
 			return &image, err
 		}
 		o.cachedImages[imageName] = image
+		return &image, nil
 	}
 
 	return &image, nil
+}
+
+func imageIsOutdated(cachedImage containerd.Image, imageUrl string) bool {
+	remoteImageDigest, err := remoteImageDigest(imageUrl)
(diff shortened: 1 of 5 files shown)


ustiugov (commented):
looks good! however, I am afraid that the image check check would slow down the system if used by default. @biecho could you please measure the added overhead? you can use timestamps printed within log messages, we just need an estimate.

biecho:
> looks good! however, I am afraid that the image check check would slow down the system if used by default. @biecho could you please measure the added overhead? you can use timestamps printed within log messages, we just need an estimate.

Done

ustiugov:
@biecho great! could you please publish the measurements in this PR? 

also, could you please take a look at the test failures (must be very simple), particularly vHive unit tests/ [misc, taps, ...]. The failure is likely due to the repo migration last week. I would really appreciate if you could help fix the problems.

biecho:
> @biecho great! could you please publish the measurements in this PR?
> 
> also, could you please take a look at the test failures (must be very simple), particularly vHive unit tests/ [misc, taps, ...]. The failure is likely due to the repo migration last week. I would really appreciate if you could help fix the problems.

Sure I'll publish the measurements as soon as possible. Regarding the vHive unit/tests, I see the following in the logs:

```

File: ctriface/iface.go [hunk 1]
```diff
@@ -314,42 +314,70 @@ func getImageURL(image string) string {
...
 		}
 		o.cachedImages[imageName] = image
+		return &image, nil
 	}
 
 	return &image, nil
 }
 
+func imageIsOutdated(cachedImage containerd.Image, imageUrl string) bool {
+	remoteImageDigest, err := remoteImageDigest(imageUrl)
```
cvetkovic, line 333: Check for err and dump the error message.

ctriface/iface.go
```diff
@@ -314,42 +314,70 @@ func getImageURL(image string) string {
...
-		} else {
-			// Pull remote image
-			image, err = o.client.Pull(ctx, imageURL,
-				containerd.WithPullUnpack,
-				containerd.WithPullSnapshotter(o.snapshotter),
-			)
-		}
+
+	imageUrl := getImageURL(imageName)
+	if !found || imageIsOutdated(image, imageUrl) {
```
cvetkovic, line 319: Can you check how image staleness is checked in `containerd`? Clone the `containerd` repo and have a look into ContainerCreate stage of the sandbox creation in code, as well as in the logs. Containerd is already running on the worker node and you can check the performance of ContainerCreate phase in containerd logs. It should be creating queue-proxy container. Look into Estella's writing to see how to find logs.

ctriface/iface.go
```diff
@@ -314,42 +314,70 @@ func getImageURL(image string) string {
...
+	return err == nil && cachedImage.Target().Digest.String() != remoteImageDigest
+}
+
+func remoteImageDigest(imageUrl string) (string, error) {
+	cmd := fmt.Sprintf("skopeo inspect docker://%s | jq -r '.Digest'", imageUrl)
+
+	start := time.Now()
+	out, err := exec.Command("bash", "-c", cmd).Output()
+	elapsed := time.Since(start)
+	log.Printf("Remote digest fetching took %s", elapsed)
```
cvetkovic, line 343: Debugf instead of Printf. If that is not available, switch to using logrus logging library.

ctriface/iface.go
```diff
@@ -314,42 +314,70 @@ func getImageURL(image string) string {
...
+
+func (o *Orchestrator) pullImage(ctx context.Context, imageUrl string) (containerd.Image, error) {
+	log.Debugf("Pulling image %s\n", imageUrl)
+
+	local, _ := isLocalDomain(imageUrl)
+
+	if local {
+		return o.pullLocalImage(ctx, imageUrl)
+	}
+	return o.pullRemoteImage(ctx, imageUrl)
```
cvetkovic, line 356: One is for the local registry and the other is for the remote one. Is that correct?

cvetkovic (changes requested):

ustiugov:
@cvetkovic do you guys still need this patch? if so, can someone take it over?

PR #628, opened 2022-10-17
Optimized dependabot config

ustiugov:
## Summary

- Disabled tracking of some benchmarks (to be decomissioned in favor of vSwarm)
- Disable dependabot updates of patch versions


dhschall (approved):
👍🏽

PR #654, opened 2022-12-29
Increase CRI test timeout

ustiugov:
## Summary

CRI timeout of 20 seconds seem too tight for executing on a self-hosted runner in a multipass VM.  

Excluded scale-from-0 test due to its sporadic failure. Raised #655


ustiugov:
Obsolete

PR #718, opened 2023-04-13
Remove rnn-serving from example functions

anshalshukla:
## Summary

rnn-serving is known to be unstable.


ustiugov:
No need to delete the YAML, other than that is ok. Thanks

PR #766, opened 2023-06-30
Make runner type configurable for CRI workflows

ustiugov:
## Summary

Make runner type configurable in the CRI tests. In particular, I would like to check if gvisor CRI can work on github-hosted runners.

* N/A

* N/A


ustiugov:
gvisor is currently tested with Istio, which has problems running in a GH-hosted runner. Hence, not going to merge it.

Issue #794, opened 2023-09-04
Move to vanilla firecracker snapshots


CuriousGeorgiy:
Currently, we have our own custom implementation of snapshots for firecracker, but it would be highly desirable to move to vanilla firecracker snapshots.

Firecracker supports snapshots since v0.23.0 [^1], they are also supported in the firecracker-go-sdk since v1.0.0 [^2]. microVM snapshots are not supported in firecracker-containerd, so we need to patch it to support snapshot-restore requests.

Firecracker snapshots have a limitation on resource names during snapshot loading, which is not compatible with container snapshot mounts, so we will need to patch it, by adding a parameter to the snapshot loading request for the new container snapshot's path. We will also need to patch the firecracker-go-sdk to forward this new request parameter. See also firecracker-microvm/firecracker#4014 for details.

[^1]: https://github.com/firecracker-microvm/firecracker/releases/tag/v0.23.0
[^2]: https://github.com/firecracker-microvm/firecracker-go-sdk/releases/tag/v1.0.0

PR #798, opened 2023-09-04
New network topology for firecracker VMs

### LINKED ISSUE #797: New network topology for firecracker VMs
Currently, each firecracker VM needs to use a TAP network device, to route its packages into the network stack of the physical host. When saving and restoring a function instance, the tap device name and the IP address of the functions’ server, running inside the container, are preserved (see also the current requirements for vanilla firecracker snapshot loading[^1]). This leads to networking conflicts on the host and limits the snapshot restoration to a single instance per physical machine. 

To bypass this obstacle, the following network topology is proposed:

Screenshot 2023-09-04 at 16 02 10

1. A new network namespace (e.g.: VMns4) is created for each VM, in which the TAP device from the snapshotted VM is rebuilt and receives the original IP address of the function. The TAP device will broadcast all the incoming and outgoing packets to and from the serverless function and VM’s network interface. Each VM will run in its own network namespace, leading to no conflicts on the host due to networking resources.

2. A local virtual tunnel is established between the VM inside its network namespace and the host node via a virtual ethernet pair (veth). A link is then established between the two ends of the virtual ethernet pair, in the network namespace (veth4-0) and the host namespace (veth4-1). In contrast, the default vHive configuration sets up a similar forwarding system through network bridges.

3. Inside the network namespace we add a routing rule that redirects all packets via the veth VM end towards a default gateway (172.17.0.17). Thus, all packets sent by the function will show at the hosts’ end of the tunnel.

4. To avoid IP conflicts when routing the packets to and from functions, each VM is assigned a unique clone address (172.18.0.5). All packets leaving the VM end of the virtual ethernet pair get their source address rewritten to the clone address of the corresponding VM. Packets entering the host end of the virtual ethernet pair get their destination address written to the original address of the VM. As a result, each VM still thinks it is using the original address while in reality, its address is translated to a clone address, different for every VM. This is accomplished using two rules in the NAT table corresponding to the virtual namespace of the VM. One rule is added in the POSTROUTING chain and one in the PREROUTING chain. The POSTROUTING rule alters the network packets before they are sent out in the virtual tunnel, from the VM namespace to the host, and rewrites the IP source address of the packet. Similarly, the PREROUTING rule overwrites the destination address of incoming packets, before routing. The two ensure that packets going into the virtual namespace have their destination address the original IP address of the VM (172.16.0.2),
while packets coming out of the namespace have their source address the clone IP address (172.18.05). The source IP address will remain the same for all the VM in the enhanced snapshotting mode, being set to 172.16.0.2 respectively.

5. In the routing table of the host, we add a rule that dictates that any package that has as destination IP the clone IP of a VM, will be routed towards the end of the tunnel situated in
the corresponding network namespace, through a set gateway (172.17.0.18). This ensures that whenever packages arrive on the host for a VM, they will be sent down the right virtual tunnel instantaneously.

6. In the hosts NFT filter table we add 2 rules for the FORWARD chain, that allow traffic from the host end of the veth pair (veth4-1) to the default host interface (eno 49) and vice versa.

The tap manager will be refactored into a new networking managing component responsible for managing the network topology described above.

[^1]: https://github.com/firecracker-microvm/firecracker/blob/main/docs/snapshotting/snapshot-support.md#loading-snapshots

CuriousGeorgiy:
## Summary

Closes #797
Part of #794

See #797 for details.

This PR introduces a new networking manager which implements the new network topology described in #797, but does not replace the existing tap manager, as it requires a patch to firecracker-containerd that adds a network namespace parameter to the CreateVM request.

Unit tests for the new networking manager are provided.

* N/A.

* N/A.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -26,7 +26,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        module: [taps, misc, profile]
+        module: [taps, misc, profile, networking]
     steps:
 
     - name: Set up Go 1.19
--- a/go.mod
+++ b/go.mod
@@ -60,6 +60,7 @@ require (
 	github.com/stretchr/testify v1.8.0
 	github.com/vhive-serverless/vhive/examples/protobuf/helloworld v0.0.0-00010101000000-000000000000
 	github.com/vishvananda/netlink v1.1.1-0.20201029203352-d40f9887b852
+	github.com/vishvananda/netns v0.0.0-20200728191858-db3c7e526aae
 	github.com/wcharczuk/go-chart v2.0.1+incompatible
 	golang.org/x/net v0.6.0
 	golang.org/x/sync v0.1.0
@@ -102,7 +103,6 @@ require (
 	github.com/opencontainers/runtime-spec v1.0.3-0.20200929063507-e6143ca7d51d // indirect
 	github.com/opencontainers/selinux v1.8.0 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
-	github.com/vishvananda/netns v0.0.0-20200728191858-db3c7e526aae // indirect
 	github.com/willf/bitset v1.1.11 // indirect
 	go.opencensus.io v0.22.4 // indirect
 	golang.org/x/image v0.7.0 // indirect
--- a/networking/Makefile
+++ b/networking/Makefile
@@ -1,0 +1,33 @@
+# MIT License
+#
+# Copyright (c) 2023 Georgiy Lebedev, Dmitrii Ustiugov, Plamen Petrov and vHive team
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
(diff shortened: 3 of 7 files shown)


ustiugov (approved):
LGTM, please check the tests

CuriousGeorgiy:
@ustiugov I made an isolated change (i.e., the new networking module is not used), so I AFAIC, the test failures are not related to my changes (and the unit test for the networking module passes).

lrq619:
I'm restarting the runners, might be related to this

CuriousGeorgiy:
@lrq619 looks like the workflows are all failing due to the same error:

```console
Error: 4m[09:32:45] [Info] Installing Knative Serving component (gvisor mode) >>>>> [09:33:34] [Error] [exit 1] -> error: unable to read URL "https://raw.githubusercontent.com/vhive-serverless/vHive/main/configs/knative_yamls/serving-core.yaml", server reported 503 Service Unavailable, status code=503
Error: 1m[09:33:[34](https://github.com/vhive-serverless/vHive/actions/runs/6074712390/job/16500708127?pr=800#step:6:35)] [Error] Failed to install Knative Serving component!
Error: 1m[09:33:34] [Error] Faild subcommand: start_onenode_vhive_cluster!
```

lrq619:
All tests passed

leokondrashov:
Merge is blocked because the commit is deemed unsigned. I think that all that novel that you put in the commit message is too much for this auto check. Can you please remove it and update the branch?

CuriousGeorgiy:
@leokondrashov No, the problem was you required signing commits for this repository. Fixed it by setting up commit signing.

Issue #807, opened 2023-09-08
Move to vanilla firecracker UPF


CuriousGeorgiy:
**Describe the enhancement**
Vanilla firecracker snapshots that will be introduced in #794 will lack UPF and thus moving to vanillla firecracker UPF will require a separate integration effort.

Support for UPF was introduced in firecracker v1.1.0[^1] and in firecracker-microvm/firecracker-go-sdk@2e09014.

[^1]: https://github.com/firecracker-microvm/firecracker/releases/tag/v1.1.0

leokondrashov:
resolved by #1221

PR #808, opened 2023-09-08
Temporarily disable UPF feature and tests on `firecracker-v1.4.1-vhive-integration` branch

CuriousGeorgiy:
## Summary

UPF feature for vanilla firecracker requires a separate integration effort (#807), so temporarily disable this feature and all tests for it.

* vHive and the vHive and orchestrator tests now exit with an error when UPF is enabled. 
* Tests using UPF are simply commented out.

* N/A.

* N/A.

### DIFF
--- a/.github/workflows/nightly_tests.yml
+++ b/.github/workflows/nightly_tests.yml
@@ -58,7 +58,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        vhive_args: ["-dbg", "-dbg -snapshots", "-dbg -snapshots -upf"]
+        # User-level page faults are temporarily disabled (gh-807)
+        # vhive_args: ["-dbg", "-dbg -snapshots", "-dbg -snapshots -upf"]
+        vhive_args: [ "-dbg", "-dbg -snapshots" ]
     env:
         GITHUB_RUN_ID: ${{ github.run_id }}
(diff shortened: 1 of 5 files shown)


ustiugov (approved):
approved. @char-1ee please take a note. You'd need to re-enable it for firecracker v1.4 hypervisor as one of the goals of your project. We need to do that ASAP. 
@BenjaminChun you also need to keep this in mind. 

For the vHive community:
Note that we don't drop support for user page faults. However, we expect them to be less stable before @char-1ee completes his task.

CuriousGeorgiy:
Please note this is not part currently part of the `main` branch, this is part of the `firecracker-v1.4.1-vhive-integration` branch, where vanilla firecracker snapshots will be introduced. 

When UPF support is added, the `firecracker-v1.4.1-vhive-integration` branch we be merged to the `main` branch.

ustiugov:
@CuriousGeorgiy I would prefer this to appear in `main` (when you are done with your changes) while having a branch `vhive-userpagefaults-legacy` for the legacy functionality users. also, the two branches need to be documented in the main README. and also document the remote snaps support limitations, i.e., what's to be done there.

We should pay attention to the new version so that it doesn't become stale. the legacy one works and no need to upgrade it. think of k8s/knative upgrade, for example

PR #809, opened 2023-09-09
redeploy cri tests nightly

lrq619:
## Summary

Add re-deployment steps to cri-runners. Now re-deployment will be run: 1. nightly 2. before every check

1. Add a new step in cri-tests: before every check, it will first re-deploy the runner
2. deployer.go now will add a cron task in 4:10am, try to re-deploy the runner nightly
3. In setup_bare_metal_runner.sh, rm the previously created service file so the re-deployment can be done without any error 

* 

* 

### DIFF
--- a/scripts/github_runner/deploy_runners.go
+++ b/scripts/github_runner/deploy_runners.go
@@ -132,8 +132,15 @@ func deployRunner(host string, runnerConf RunnerConf, deployerConf *DeployerConf
 	var setupCmd string
 	switch runnerConf.Type {
 	case "cri":
+		log.Debugf("Adding redploy crontab task to %s@%s", deployerConf.HostUsername, host)
 		setupCmd = fmt.Sprintf("cd vhive && ./scripts/github_runner/setup_bare_metal_runner.sh %s %s %s", deployerConf.GhOrg,
 			deployerConf.GhPat, runnerConf.Sandbox)
+		var redeploySetupCmd string = fmt.Sprintf("echo '10 4 * * * root rm -rf ./runners/ && %s' >> etc/crontab", setupCmd)
+		out, err = client.Exec(redeploySetupCmd)
+		log.Debug(string(out))
+		if err != nil {
+			log.Fatalf("Failed to setup redeploy task on %s@%s: %s", deployerConf.HostUsername, host, err)
+		}
 	case "integ":
 		var restart string
 		if runnerConf.Restart {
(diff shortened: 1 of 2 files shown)


scripts/github_runner/setup_bare_metal_runner.sh
```diff
@@ -37,6 +37,7 @@ GH_PAT=$2
 SANDBOX=$3
 
 VHIVE_ROOT="$(git rev-parse --show-toplevel)"
+rm /etc/systemd/system/actions.runner.vhive-serverless-vhive.*.service
```
ustiugov, line 40:
```suggestion
rm /etc/systemd/system/actions.runner.vhive-serverless-vhive.*.service || echo No need to clean the node
```

provision for the clean node otherwise the script would fail

ustiugov (changes requested):
Relevant? otherwise please close.

lrq619:
I removed the re-deployment in cri-test.yml, add the condition check for rm service files

scripts/github_runner/setup_bare_metal_runner.sh
```diff
@@ -36,6 +36,16 @@ GH_ORG=$1
 GH_PAT=$2
 SANDBOX=$3
 
+# clean the service file before proceeding
+pattern="/etc/systemd/system/actions.runner.vhive-serverless-vhive.*.service"
+if ls $pattern 1> /dev/null 2>&1; then
+  rm $pattern
+  echo "Deleted matching files."
```
ustiugov, line 43: be specific which files, i.e., belong to what entity

scripts/github_runner/deploy_runners.go
```diff
@@ -134,6 +134,12 @@ func deployRunner(host string, runnerConf RunnerConf, deployerConf *DeployerConf
 	case "cri":
 		setupCmd = fmt.Sprintf("cd vhive && ./scripts/github_runner/setup_bare_metal_runner.sh %s %s %s", deployerConf.GhOrg,
 			deployerConf.GhPat, runnerConf.Sandbox)
+		var redeploySetupCmd string = fmt.Sprintf("echo '10 4 * * * root rm -rf ./runners/ && %s' >> etc/crontab", setupCmd)
+		out, err = client.Exec(redeploySetupCmd)
+		log.Debug(string(out))
```
ustiugov, line 139: need more elaborate output

ustiugov (changes requested):

scripts/github_runner/deploy_runners.go
lrq619: Add a line of output indicating what command will be executed

scripts/github_runner/setup_bare_metal_runner.sh
lrq619: Done

ustiugov (approved):

PR #811, opened 2023-09-11
Encapsulate the host interface name into the VM pool and tap manager

### LINKED ISSUE #810: Encapsulate the host interface name into the VM pool and tap manager
**Describe the enhancement**
Currently, the host interface name is separated from the VM pool and tap manager, and as a result, it needs to be passed to all VM pool methods.

This is inconvenient, and it seems that the host interface name should be encapsulated into the VM pool and tap manager.

CuriousGeorgiy:
#### Encapsulate the host interface name into the VM pool and tap manager

Currently, the host interface name is separated from the VM pool and tap
manager, and as a result, it needs to be passed to all VM pool methods.

This is inconvenient, and so encapsulate the host interface name into the
VM pool and tap manager.

Closes #810
Part of #794

* The host interface name is encapsulated into the VM pool and tap manager.

* N/A.

* N/A.

### DIFF
--- a/cri/firecracker/service.go
+++ b/cri/firecracker/service.go
@@ -216,5 +216,5 @@ func getEnvVal(key string, config *criapi.ContainerConfig) (string, error) {
 
 	}
 
-	return "", errors.New("failed to provide non empty guest image in user container config")
+	return "", errors.New("failed to retrieve environment variable from user container config")
 }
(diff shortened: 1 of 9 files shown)


ustiugov (approved):
thx. please merge when tests pass.

PR #816, opened 2023-09-11
Move to vanilla firecracker snapshots

CuriousGeorgiy:
## Summary

Closes #794

## Implementation Notes :hammer_and_pick:

* Replaced the tap manager with the new network manager and dropped the tap manager;
* Reworked snapshot manager to maintain one list of snapshots with a new `ready` feature;
* Created a device snapshot for the base container image during patch creation;
* Dropped VM offloading APIs and replaced them with snapshot creation or VM shutdown;
* Added a new network pool size option to the orchestrator;
* Updated firecracker-related binaries;
* Updated firecracker integration code;
* Added basic remote snapshot create and load tests.

### Snapshot revisions

Snapshots are now identified by a function instance _revision_, which is naturally provided by the `K_REVISION` environment variable of knative[^1].

## External Dependencies :four_leaf_clover:

* #811 must be merged first;
* #817 must be merged first;
* Added rsync to system dependencies.

## Breaking API Changes :warning:

* N/A.

[^1]: https://knative.dev/docs/concepts/serving-resources/revisions/

PR #817, opened 2023-09-12
Various enhancements to the vHive function pool

CuriousGeorgiy:
### Propagate `StartVMResponse` from `Orch:LoadSnapshot` to `Func:LoadInstance`

This will come in handy when the loaded instance will be a snapshot booted
VM rather than an offloaded one.

### Pass VM ID of function by value in goroutine for async instance removal

Goroutines capture variables by reference, so it is safer to make a copy
(i.e., a local variable) and pass the variable by value.

This will also come in handy when the function's VM ID can change when it
is loaded from a snapshot booted VM rather than an offloaded one.

### Pass VM ID to `Function:LoadInstance` instead of relying on existing one

This will come in handy when the function's VM ID will be assigned to
it after a VM for it is loaded from a snapshot (i.e., a new VM is created
rather than an offloaded one is reused). This logic follows the branch when
a snapshot is not available and a VM is booted from scratch.

* Replace manually created snapshots with snapshot manager in function pool;
* Propagate `StartVMResponse` from `Orch:LoadSnapshot` to `Func:LoadInstance`;
* Pass VM ID of function by value in goroutine for async instance removal (`RemoveInstanceAsync`);
* Pass VM ID to `Function:LoadInstance` instead of relying on existing one.

* N/A.

* N/A.

### DIFF
--- a/functions.go
+++ b/functions.go
@@ -25,6 +25,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"github.com/vhive-serverless/vhive/ctriface"
 	"math/rand"
 	"net"
 	"os"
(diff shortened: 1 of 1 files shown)


ustiugov (dismissed):
lgtm

CuriousGeorgiy:
@ustiugov turned out the commit that starts using the snapshot manager in the function pool doesn't work with the offload API, so I dropped it from this patch set and squashed it with the main commit of #816.

Please re-review.

lrq619:
Checks all passed

ustiugov (approved):

PR #878, opened 2023-11-09
change to utest runner

lrq619:
## Summary

PMU tools can not work normally on github hosted runners
Add a self hosted runner and assign profile unit test running on that runner

* Change the implementation of detecting pmu tools and perf: stdout cannot be captured in self-hosted runners

* N/A

* N/A

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -26,7 +26,52 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        module: [misc, profile, networking, snapshotting]
+        module: [misc, networking, snapshotting]
+    steps:
+
+    - name: Set up Go 1.19
+      uses: actions/setup-go@v4
+      with:
+        go-version: '1.19'
+        
+    - name: Set up Python 3.x
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.x'
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v4
+
+    - name: Build setup scripts
+      run: pushd scripts && go build -o setup_tool && popd
+      
+    - name: Add rsync
+      run: |
+        sudo apt update
+        sudo apt install rsync -y
+
+    - name: Setup System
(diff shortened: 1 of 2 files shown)


profile/profiler.go
```diff
@@ -298,23 +298,24 @@ func (p *Profiler) parseMetric(lines []pmuLine) map[string]float64 {
...
+	return true
 }
 
 func isPerfInstalled() bool {
-	cmd := exec.Command("perf", "--version")
-	b, err := cmd.Output()
+	_, err := exec.LookPath("perf")
 	if err != nil {
-		log.Error(err)
+		log.Printf("perf is not installed or not found in PATH: %v", err)
```
leokondrashov, line 315: Use `Error` for that

.github/workflows/unit_tests.yml
```diff
@@ -22,7 +22,7 @@ env:
 jobs:
   unit-test:
     name: Unit test
-    runs-on: ubuntu-20.04
+    runs-on: [self-hosted, Linux, X64, utest]
```
leokondrashov, line 25: Do we need to move all of the unit tests on self-hosted?

leokondrashov (changes requested):

profile/profiler.go
lrq619: OK, will change to `Errorf`

.github/workflows/unit_tests.yml
lrq619: I think so, all jobs under `unit-test` include the step `Install PMU tools`

leokondrashov: Yeah, but only the profile unit test actually needs it. You've tried that in separate PR. I'm thinking about offloading to self-hosted only the parts that cannot be ran on github runners. Just like firecracker unit tests in the same file.

lrq619: Okay, no problem. I will change the yaml structure

leokondrashov (approved):
lgtm

PR #886, opened 2023-11-17
Openyurt nodes deployment configs

jchua99:
## Summary

Config nodes to setup kube cluster, knative. Config clode and edge nodepool with template

* ./openyurtdeployer deploy
    * read and unmarshall json
    * load system, knative, kube config from vhive/configs/setup
    * stores data in nodeList to use to deploy

Folders
* configs - contains config struct for setup
* node - contain functions to run command to setup 
* template - yaml template
 
* 

* 


leokondrashov:
@jchua99 Please create more clear names for PRs. 'replay config part' doesn't even make any sence to anyone opening list of PRs.

lrq619:
It's generally good. But be aware of following points:

- [ ] Enable CI workflow in `openyurt` branch
- [ ] Golang functions can return multiple values, and the common way is to return a result and an `err`, and handle this `err` mannually in the callee side of this function
- [ ] Some config files can be reused, they are similar to files in `https://github.com/vhive-serverless/vHive/tree/main/scripts/configs`, please carefully examine the difference and similarity of them and try to reuse the config files as much as possible

In this branch, we have setup tool to directly setup a k8s cluster and knative cluster, which is a useful tool that needs to be merged into the `setup` tool in the main branch, but this can be done refactoring `openyurt`, just be aware of the possible future task

lrq619:
Jason, please always remember to squash all commits into one after every git commit. Please do that now.

PR #888, opened 2023-11-18
Re-structure knative and istio YAML usage

wanghanchengchn:
Re-structure knative and istio YAML usage

## Summary

Previously, we fetched the Knative and net-istio YAML files for container-based deployments from the Internet. However, a large number of applications, such as loaders, need to use modified YAML files. Therefore, I modified the code for vhive to apply YAML files so that vhive can use customized YAML files.

## Implementation Notes :

* Specifically, I modified the code in scripts/cluster/*.go to apply the YAML files. 
* In addition, I also modified the IP address in configs/setup/kube.json.


scripts/cluster/setup_master_node.go
```diff
@@ -199,7 +214,13 @@ func InstallKnativeServingComponent(stockContainerd string) error {
 		if !utils.CheckErrorWithMsg(err, "Failed to install Knative Serving component!\n") {
 			return err
 		}
-		_, err = utils.ExecShellCmd("kubectl apply -f https://github.com/knative/serving/releases/download/knative-v%s/serving-core.yaml", configs.Knative.KnativeVersion)
+		
+		loaderServingCorePath, err := utils.GetVHiveFilePath(path.Join("configs/knative_yamls", "loader_serving_core.yaml"))
```
leokondrashov, line 218: Please rename these files to be clearer what each one does: we have one for containerd (with stock version of knative images) and for firecracker (with custom versions).

In the end, there should be no `loader` in the names of these files.

scripts/cluster/setup_master_node.go
```diff
@@ -263,7 +290,14 @@ func ConfigureMagicDNS() error {
 // Deploy Istio pods
 func DeployIstioPods() error {
 	utils.WaitPrintf("Deploying istio pods")
-	_, err := utils.ExecShellCmd("kubectl apply -f https://github.com/knative/net-istio/releases/download/knative-v%s/net-istio.yaml", configs.Knative.KnativeVersion)
+	
+	loaderIstioControllerPath, err := utils.GetVHiveFilePath(path.Join("configs/knative_yamls", "loader_istio_controller.yaml"))
```
leokondrashov, line 294: This file is not found in the main vhive branch.

I propose search for this file in the repo, in case it is not found, download and use the default one. Same goes for the serving-core stuff. In stock-only case, running these scripts __not__ from the loader should install default yamls.

scripts/cluster/setup_worker_kubelet.go
```diff
@@ -51,7 +51,7 @@ func CreateWorkerKubeletService(criSock string) error {
 	}
 	bashCmd := "sudo sh -c 'cat <<EOF > /etc/systemd/system/kubelet.service.d/0-containerd.conf\n" +
 		"[Service]\n" +
-		`Environment="KUBELET_EXTRA_ARGS=--container-runtime=remote --runtime-request-timeout=15m --container-runtime-endpoint=unix://%s"` +
+		`Environment="KUBELET_EXTRA_ARGS=--container-runtime=remote --v=5 --kube-api-qps=50000 --kube-api-burst=100000 --runtime-request-timeout=15m --container-runtime-endpoint=unix://%s"` +
```
leokondrashov, line 54: @cvetkovic Do you think these API qps changes are good-to-have everytime running vHive? Can they be applied on demand based on the needed values for experiment?

leokondrashov (changes requested):

configs/setup/kube.json
```diff
@@ -2,7 +2,7 @@
 	"K8sVersion": "1.25.9",
 	"AlternativeImageRepo": "",
 	"ApiserverAdvertiseAddress": "",
-	"PodNetworkCidr": "192.168.0.0/16",
+	"PodNetworkCidr": "10.168.0.0/16",
```
leokondrashov, line 5: Can you also try to change these addresses on loader side? So we won't need to change them here.

scripts/cluster/setup_master_node.go
```diff
@@ -199,18 +215,42 @@ func InstallKnativeServingComponent(stockContainerd string) error {
 		if !utils.CheckErrorWithMsg(err, "Failed to install Knative Serving component!\n") {
 			return err
 		}
-		_, err = utils.ExecShellCmd("kubectl apply -f https://github.com/knative/serving/releases/download/knative-v%s/serving-core.yaml", configs.Knative.KnativeVersion)
-		if !utils.CheckErrorWithTagAndMsg(err, "Failed to install Knative Serving component!\n") {
-			return err
+		
+		if _, fileStat := os.Stat(path.Join(configs.VHive.VHiveRepoPath, path.Join("configs/knative_yamls", "container-serving-core.yaml"))); fileStat != nil {
```
leokondrashov, line 219: Didn't follow the logic: if we have valid fileStat we install from the Internet, right? Sounds wrong, should be the opposite.

scripts/cluster/setup_master_node.go
```diff
@@ -199,18 +215,42 @@ func InstallKnativeServingComponent(stockContainerd string) error {
...
 		_, err := utils.ExecShellCmd("kubectl apply -f %s/serving-crds.yaml", configs.Knative.NotStockOnlyKnativeServingYamlUrlPrefix)
 		if !utils.CheckErrorWithMsg(err, "Failed to install Knative Serving component!\n") {
 			return err
 		}
-		_, err = utils.ExecShellCmd("kubectl apply -f %s/serving-core.yaml", configs.Knative.NotStockOnlyKnativeServingYamlUrlPrefix)
-		if !utils.CheckErrorWithTagAndMsg(err, "Failed to install Knative Serving component!\n") {
-			return err
+
+		if _, fileStat := os.Stat(path.Join(configs.VHive.VHiveRepoPath, path.Join("configs/knative_yamls", "firecracker-serving-core.yaml"))); fileStat != nil {
+			_, err = utils.ExecShellCmd("kubectl apply -f %s/serving-core.yaml", configs.Knative.NotStockOnlyKnativeServingYamlUrlPrefix)
```
leokondrashov, line 241: I propose different naming scheme: serving-core.yaml for containerd and serving-core-firecracker.yaml for firecracker version, no matter where it comes from. It should be renamed in vhive repo and the same convention used for rewriting them with yq in loader.

scripts/cluster/setup_master_node.go
```diff
@@ -199,18 +215,42 @@ func InstallKnativeServingComponent(stockContainerd string) error {
...
+			if err != nil {
+				return err
+			}
+			_, err = utils.ExecShellCmd("cat %s | envsubst | kubectl apply -f -", servingCorePath)
+			if !utils.CheckErrorWithTagAndMsg(err, "Failed to install Knative Serving component!\n") {
+				return err
+			}
 		}
 	} else {
 		_, err := utils.ExecShellCmd("kubectl apply -f %s/serving-crds.yaml", configs.Knative.NotStockOnlyKnativeServingYamlUrlPrefix)
```
leokondrashov, line 235: Same, rename this file, since it is made for firecracker.

leokondrashov (changes requested):

scripts/cluster/setup_master_node.go
```diff
@@ -199,18 +215,42 @@ func InstallKnativeServingComponent(stockContainerd string) error {
...
 		if !utils.CheckErrorWithMsg(err, "Failed to install Knative Serving component!\n") {
 			return err
 		}
-		_, err = utils.ExecShellCmd("kubectl apply -f %s/serving-core.yaml", configs.Knative.NotStockOnlyKnativeServingYamlUrlPrefix)
-		if !utils.CheckErrorWithTagAndMsg(err, "Failed to install Knative Serving component!\n") {
-			return err
+
+		if _, err = os.Stat(path.Join(configs.VHive.VHiveRepoPath, path.Join("configs/knative_yamls", "serving-core-firecracker.yaml"))); err != nil {
```
leokondrashov, line 240: This if is redundant, we always should find this file in the folder. I think, you should remove completely the field NotStockOnlyKnativeServingYamlUrlPrefix from configs then.

scripts/cluster/setup_master_node.go
leokondrashov: Same here, rename it to follow our convention of naming yamls based on function (firecracker here). And make it using local file instead of downloading it from github.

scripts/cluster/setup_master_node.go
```diff
@@ -199,18 +215,42 @@ func InstallKnativeServingComponent(stockContainerd string) error {
 		if !utils.CheckErrorWithMsg(err, "Failed to install Knative Serving component!\n") {
 			return err
 		}
-		_, err = utils.ExecShellCmd("kubectl apply -f https://github.com/knative/serving/releases/download/knative-v%s/serving-core.yaml", configs.Knative.KnativeVersion)
-		if !utils.CheckErrorWithTagAndMsg(err, "Failed to install Knative Serving component!\n") {
-			return err
+		
+		if _, err = os.Stat(path.Join(configs.VHive.VHiveRepoPath, path.Join("configs/knative_yamls", "serving-core.yaml"))); err != nil {
+			_, err = utils.ExecShellCmd("kubectl apply -f https://github.com/knative/serving/releases/download/knative-v%s/serving-core.yaml", configs.Knative.KnativeVersion)
```
leokondrashov, line 220: I think you should add a print saying that 'Found local serving-core.yaml. Using it isntead of stock version of knative'.

scripts/cluster/setup_master_node.go
```diff
@@ -199,18 +215,42 @@ func InstallKnativeServingComponent(stockContainerd string) error {
...
+				return err
+			}
+			_, err = utils.ExecShellCmd("cat %s | envsubst | kubectl apply -f -", servingCorePath)
+			if !utils.CheckErrorWithTagAndMsg(err, "Failed to install Knative Serving component!\n") {
+				return err
+			}
 		}
 	} else {
-		_, err := utils.ExecShellCmd("kubectl apply -f %s/serving-crds.yaml", configs.Knative.NotStockOnlyKnativeServingYamlUrlPrefix)
+		_, err := utils.ExecShellCmd("kubectl apply -f %s/serving-crds-firecracker.yaml", configs.Knative.NotStockOnlyKnativeServingYamlUrlPrefix)
```
leokondrashov, line 235: Same here, make it using local version.

leokondrashov (changes requested):

scripts/cluster/setup_worker_kubelet.go
cvetkovic: These values were experiment-specific.

leokondrashov: So should we remove them then?

cvetkovic: The default values are 5 and 10, I think. I would put 50 and 100 and that would be it.

leokondrashov: @wanghanchengchn please implement

scripts/cluster/create_multinode_cluster.go
```diff
@@ -69,7 +74,7 @@ func CreateMasterKubeletService() error {
 	utils.WaitPrintf("Creating kubelet service")
 	bashCmd := `sudo sh -c 'cat <<EOF > /etc/systemd/system/kubelet.service.d/0-containerd.conf
 [Service]                                                 
-Environment="KUBELET_EXTRA_ARGS=--container-runtime=remote --runtime-request-timeout=15m --container-runtime-endpoint=unix:///run/containerd/containerd.sock"
+Environment="KUBELET_EXTRA_ARGS=--container-runtime=remote --v=5 --runtime-request-timeout=15m --container-runtime-endpoint=unix:///run/containerd/containerd.sock"
```
ustiugov, line 77: I think we should consider hooking this `--v` to the debug option (do we have one already?). we don't want too much log activity when we run our system under high load.

ustiugov (changes requested):
- [ ] hook up kubectl's `--v` option to the app's `-dbg` runtime argument. if it doesn't exist, please create one

Probably not in this PR, but need to do so when the YQ-based YAML rewriting will be ready.
- [ ] generate Firecracker yamls from the original ones using YQ instead of keeping a copy of each file.
- [ ] replace `envsubst` with YQ queries

scripts/cluster/setup_master_node.go
```diff
@@ -199,20 +204,44 @@ func InstallKnativeServingComponent(stockContainerd string) error {
...
 		if !utils.CheckErrorWithMsg(err, "Failed to install Knative Serving component!\n") {
 			return err
 		}
-		_, err = utils.ExecShellCmd("kubectl apply -f %s/serving-core.yaml", configs.Knative.NotStockOnlyKnativeServingYamlUrlPrefix)
+
+		servingCorePath, err := utils.GetVHiveFilePath(path.Join("configs/knative_yamls", "serving-core-firecracker.yaml"))
+		if err != nil {
+			return err
+		}
+		_, err = utils.ExecShellCmd("cat %s | envsubst | kubectl apply -f -", servingCorePath)
```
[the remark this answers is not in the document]
leokondrashov: I think we already moved everything there, so this should be easy to just remove

scripts/cluster/create_multinode_cluster.go
leokondrashov: We have it only for vhive binary, used only in single node cluster. We use it for CI runners mostly.

I think we can add it to configs/setup/system.json for using it dusring setup. Please don't forget to add the description of new debug option to documentation (at least, to docs/developers_guide.md)

docs/developers_guide.md
```diff
@@ -318,3 +318,7 @@ Wed May 24 02:04:40 2023
 +---------------------------------------------------------------------------------------+
 
 ```
+
+## Debug Mode
+
+When `DebugMode` in `configs/setup/system.json` is set to true, we add the `--v=5` parameter to set the log level verbosity of kubelet to 5. In addition, we also execute the `IncreaseLogSizePerContainer` function to increase the LogSize of the Container. The default value of `DebugMode` is false.
```
leokondrashov, line 324: I'd say, make it more high level, like "In order to simplify the debugging procedure of whole created cluster, `DebugMode` parameter in `configs/setup/system.json` can be set to true. This debug mode increases the amount of log messages from the kubelet and increases the maximum size of stored logs from containers. By default, this mode is disabled for performance reasons."

leokondrashov (changes requested):

configs/setup/system.json
```diff
@@ -13,5 +13,6 @@
 	"KubeVersion": "1.25.9-00",
 	"PmuToolsRepoUrl": "https://github.com/vhive-serverless/pmu-tools",
 	"ProtocVersion": "3.19.4",
-	"ProtocDownloadUrlTemplate": "https://github.com/protocolbuffers/protobuf/releases/download/v%s/protoc-%s-linux-x86_64.zip"
+	"ProtocDownloadUrlTemplate": "https://github.com/protocolbuffers/protobuf/releases/download/v%s/protoc-%s-linux-x86_64.zip",
+    "DebugMode": false
```
ustiugov, line 17: fix the indent please

name it ```"LogVerbosity"```, should be 0 by default (but in CI better set to a high value). then you can directly use this value for setting verbosity, see below. people can increase verbosity by changing this value

docs/developers_guide.md
```diff
@@ -318,3 +318,7 @@ Wed May 24 02:04:40 2023
 +---------------------------------------------------------------------------------------+
 
 ```
+
```
ustiugov, line 321: I think this section should much higher up in this doc. Also, I don't think it deserves an entire section. just find the appropriate place in this doc

scripts/cluster/create_multinode_cluster.go
```diff
@@ -87,7 +99,15 @@ EOF'`
 func DeployKubernetes() error {
 
 	utils.WaitPrintf("Deploying Kubernetes(version %s)", configs.Kube.K8sVersion)
-	shellCmd := fmt.Sprintf("sudo kubeadm init --kubernetes-version %s --pod-network-cidr=\"%s\" ", configs.Kube.K8sVersion, configs.Kube.PodNetworkCidr)
+	masterNodeIp, iperr := utils.ExecShellCmd(`ip route | awk '{print $(NF)}' | awk '/^10\..*/'`)
+	if iperr != nil {
+		return iperr
+	}
+	var verbose string
+	if configs.System.DebugMode {
```
ustiugov, line 107: same fix here

scripts/cluster/create_multinode_cluster.go
```diff
@@ -67,10 +74,15 @@ func CreateMultinodeCluster(stockContainerd string) error {
...
-EOF'`
+
+	var verbose string
+	if configs.System.DebugMode {
+		verbose = "--v=5"
+	}
+	bashCmd := fmt.Sprintf(`sudo sh -c 'cat <<EOF > /etc/systemd/system/kubelet.service.d/0-containerd.conf
+		[Service]
+		Environment="KUBELET_EXTRA_ARGS=--container-runtime=remote %s --runtime-request-timeout=15m --container-runtime-endpoint=unix:///run/containerd/containerd.sock"
+		EOF'`, verbose)
```
ustiugov, line 85:
```suggestion
	bashCmd := fmt.Sprintf(`sudo sh -c 'cat <<EOF > /etc/systemd/system/kubelet.service.d/0-containerd.conf
		[Service]
		Environment="KUBELET_EXTRA_ARGS=--container-runtime=remote --v=%s --runtime-request-timeout=15m --container-runtime-endpoint=unix:///run/containerd/containerd.sock"
		EOF'`, configs.System.DebugMode)
```

scripts/cluster/setup_worker_kubelet.go
```diff
@@ -49,11 +50,16 @@ func CreateWorkerKubeletService(criSock string) error {
 	if !utils.CheckErrorWithMsg(err, "Failed to create kubelet service!\n") {
 		return err
 	}
+	var verbose string
+	if configs.System.DebugMode {
+		verbose = "--v=5"
```
ustiugov, line 55: same here

ustiugov (changes requested):
thx, almost there

docs/developers_guide.md
leokondrashov: There is a logging.md file. I think that's a good place for that, if we are renaming it to LogVerbosity.

docs/logging.md
```diff
@@ -20,7 +20,10 @@ We present how to set up a multi-node cluster, however, the same modifications c
    > **Beware:**
    > The following commands are meant to be run in a shell of type `bash` as mentioned in the [Quickstart guide][shell-type].
    > Running them with a different kind of shell interpreter can lead to undefined behaviour.
-   >
+   > 
+   > In addition, in order to simplify the debugging procedure of whole created cluster, `LogVerbosity` parameter in `configs/setup/system.json` can be set to different integers (e.g. 5).
```
leokondrashov, line 24: I think, it should be placed as usual paragraph before the notes in this section. Starting with the fact that kubelet has its own logs, that can have different verbosity. This verbosity can be controlled by this option. In addition it changes the sizes of per-container logs. And the default value.

leokondrashov (approved):

leokondrashov (approved):

PR #892, opened 2023-11-24
Github work flow to deploy configs 

jchua99:
## Summary

Workflow set up golang, builds and verify build with ./openyurt_deployer help.
Next step to set up ssh service, keygen, copy the public key into auth key, add the the ssh-agent, restart the ssh daemon, before running the script.

* Briefly outline the overall technical solution. If necessary, identify talking points where the reviewer's attention should be drawn to.

* 

* 


lrq619:
This shouldn't be a separate PR, it should be part of https://github.com/vhive-serverless/vHive/pull/887
Also, please check the `Set up SSH and run deploy` step in the workflow, it actually fails but the test still gets passed. This is not expected.

PR #897, opened 2023-12-07
hotfix to make vhive work on new package repository

JooyoungPark73:
WARN: will break all features except stock-only; opened issue to k8s

## Summary
Before migrating to `pkgs.k8s.io`, `10-kubeadm.conf` was located in `/etc/systemd/system/kubelet.service.d`
After Migrating, `10-kubeadm.conf` is located in `/usr/lib/systemd/system/kubelet.service.d/`

https://github.com/kubernetes/release/issues/3392
Currently opened an issue.
It seems that /usr/lib is default option for centOS and /etc is default on Ubuntu/Debian, and after migration, it seems that they removed this per-OS option (take a look at the conversation above)

### DIFF
--- a/scripts/cluster/create_multinode_cluster.go
+++ b/scripts/cluster/create_multinode_cluster.go
@@ -74,7 +74,7 @@ func CreateMultinodeCluster(stockContainerd string) error {
 // Create kubelet service on master node
 func CreateMasterKubeletService() error {
 	utils.WaitPrintf("Creating kubelet service")
-	bashCmd := `sudo sh -c 'cat <<EOF > /etc/systemd/system/kubelet.service.d/0-containerd.conf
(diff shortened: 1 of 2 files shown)


ustiugov:
@JooyoungPark73 is the PR ready or it draft? you may want to ask Leonid for a review

JooyoungPark73:
@ustiugov I left this a draft because k8s apt repo maintainers are currently investigating this issue, and their future direction is uncertain.

leokondrashov:
I think, if this PR fixes at least partially the problem with the k8s repos, we should merge it. Because right now, it doesn't work at all.

leokondrashov (approved):
K8s folks are proposing different way of dropping in the arguments. We'll merge it for now, so it works, and open an issue about a proper way of doing that.

PR #928, opened 2024-02-01
Firecracker snapshot upf ctriface

char-1ee:
This is an informal PR for debugging. Please DO NOT merge.

1. These updates are based on my `firecracker-snapshot-upf-test` branch: https://github.com/char-1ee/vHive/tree/firecracker-snapshot-upf-test. 
2. These updates currently cannot pass integration tests in `./vhive_test.go`
3. I committed the updates that can pass `./ctriface/iface_test.go`, which contains the `TestStartSnapStop()` from `failing_test.go`.


char-1ee:
@ustiugov

ctriface/iface.go
```diff
@@ -64,6 +65,8 @@ type StartVMResponse struct {
 
 const (
 	testImageName = "ghcr.io/ease-lab/helloworld:var_workload"
+	fileBackend   = "File"
```
ustiugov, line 68: please don't use hardcoded values in the modules. have all the hardcoded values in the test's code

ctriface/iface.go
```diff
@@ -508,7 +523,8 @@ func (o *Orchestrator) LoadSnapshot(ctx context.Context, vmID string, snap *snap
 	go func() {
 		defer close(loadDone)
 
-		if _, loadErr = o.fcClient.CreateVM(ctx, conf); loadErr != nil {
+		_, loadErr := o.fcClient.CreateVM(ctx, conf)
+		if loadErr != nil {
```
ustiugov, line 527: this is the same, revert

ctriface/iface.go
```diff
@@ -536,7 +552,25 @@ func (o *Orchestrator) LoadSnapshot(ctx context.Context, vmID string, snap *snap
...
+			VMID:             vmID,
+			GuestMemPath:     o.getMemoryFile(vmID),
+			BaseDir:          o.getVMBaseDir(vmID),
+			GuestMemSize:     int(conf.MachineCfg.MemSizeMib) * 1024 * 1024,
+			IsLazyMode:       o.isLazyMode,
+			VMMStatePath:     o.getSnapshotFile(vmID),
+			WorkingSetPath:   o.getWorkingSetFile(vmID),
+			InstanceSockAddr: o.uffdSockAddr,
+		}
+		if err := o.memoryManager.RegisterVM(stateCfg); err != nil {
```
ustiugov, line 570: hmm I am surprised that registration was not there. I am quite sure it is/was done inside the memory manager

ctriface/iface.go
```diff
@@ -205,17 +208,18 @@ func (o *Orchestrator) StartVMWithEnvironment(ctx context.Context, vmID, imageNa
...
+			GuestMemPath:     o.getMemoryFile(vmID),
+			BaseDir:          o.getVMBaseDir(vmID),
+			GuestMemSize:     int(conf.MachineCfg.MemSizeMib) * 1024 * 1024,
+			IsLazyMode:       o.isLazyMode,
+			VMMStatePath:     o.getSnapshotFile(vmID),
+			WorkingSetPath:   o.getWorkingSetFile(vmID),
+			InstanceSockAddr: o.uffdSockAddr,
 		}
+		logger.Debugf("TEST: show to-reg snapStat: %+v", stateCfg)
 		if err := o.memoryManager.RegisterVM(stateCfg); err != nil {
```
ustiugov, line 223: ok, VM registration happens here. You should not reuse/repurpose this interface for LoadSnapshot. If needed, create a new interface.

so we would have 2 interfaces for registration: one for the original VMs and one fo the VMs which are loaded from a snap.

ctriface/iface.go
```diff
@@ -33,6 +32,8 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/vhive-serverless/vhive/snapshotting"
```
ustiugov, line 35: this doesn't look correct to me. memory manager should be the one that includes this module

ctriface/manual_cleanup_test.go
```diff
@@ -91,7 +91,7 @@ func TestSnapLoad(t *testing.T) {
 
 	vmID = "2"
 
-	_, _, err = orch.LoadSnapshot(ctx, vmID, snap)
+	_, _, err = orch.LoadSnapshot(ctx, "1", vmID, snap)
```
ustiugov, line 94: please replace contstants with a properly names variable for readability. In this case, can use `originalVmID`

ctriface/orch.go
```diff
@@ -121,8 +122,16 @@ func NewOrchestrator(snapshotter, hostIface string, opts ...OrchestratorOption)
 	}
 
 	if o.GetUPFEnabled() {
+		// o.uffdSockAddr = "/tmp/uffd.sock" 	// "/tmp/uffd/firecracker-containerd#3-0/uffd.sock"
+		_, err = os.Create("/tmp/uffd.sock")
```
ustiugov, line 126: don't use hardcoded values. pass these values as arguments from tests

cri/firecracker/coordinator.go
```diff
@@ -181,7 +182,7 @@ func (c *coordinator) orchLoadInstance(ctx context.Context, snap *snapshotting.S
 	ctxTimeout, cancel := context.WithTimeout(ctx, time.Second*30)
 	defer cancel()
 
-	resp, _, err := c.orch.LoadSnapshot(ctxTimeout, vmID, snap)
+	resp, _, err := c.orch.LoadSnapshot(ctxTimeout, vmID, vmID, snap)
```
ustiugov, line 185: bad naming of the arguments

ustiugov (commented):
most importantly, please fix the naming

ctriface/iface.go
char-1ee: These 2 const (fileBackend, uffdBackend) should be independent of the test. If UPF enabled, the code pass uffBackend into CreateVM(),  else, the code pass fileBackend into CreateVM().

ctriface/iface.go
char-1ee: this module is imported because `snap *snapshotting.Snapshot` is used in several function parameters in iface.go.

leokondrashov:
superceeded by #1221

PR #936, opened 2024-02-12
Fix firecracker related issues

### LINKED ISSUE #694: required key GUEST_ADDR missing value
when i deploy service by file yaml:
apiVersion: serving.knative.dev/v1
kind: Service
metadata:
  name: aes-python
  namespace: default
spec:
  template:
    spec:
      containers:
        - image: crccheck/hello-world:latest # Stub image. See https://github.com/ease-lab/vhive/issues/68
          ports:
            - name: h2c # For GRPC support
              containerPort: 50051
          env:
            - name: GUEST_PORT # Port on which the firecracker-containerd container is accepting requests
              value: "50051"
            - name: GUEST_IMAGE # Container image to use for firecracker-containerd container
              value: "docker.io/vhiveease/aes-python:latest"
It show error: required key GUEST_ADDR missing value
How to fix error?

### LINKED ISSUE #899: Properly setup K8S services
From handling #898 , the k8s repo maintainers pointed out that our implementation is not compatible.

https://github.com/kubernetes/release/issues/3392#issuecomment-1849684117
We should not use `KUBELET_EXTRA_ARGS` but use kubeadm configuration files.

I think we should refrain from using this drop-in, so this issue should be addressed in the near future.

### LINKED ISSUE #931: Error while deploying firecracker single-node deployment
on previous script implementation, https://github.com/vhive-serverless/vHive/blob/69c5975eb30a4a2fa661559cdedce31ce2a70ebf/scripts/setup_firecracker_containerd.sh#L49

stored vmlinux as `hello-vmlinux.bin` , but now we just store it as `vmlinux-5.10.186`

https://github.com/vhive-serverless/vHive/blob/69c5975eb30a4a2fa661559cdedce31ce2a70ebf/scripts/setup/setup.go#L87-L94

need to rename before moving.

JooyoungPark73:
Fixes #931 **Error while deploying firecracker single-node deployment**
This issue is caused when migrating from bash script to go script. Did not change name properly.
https://github.com/vhive-serverless/vHive/blob/14f0b2d7ca9ab2a243063335c996e79d70349c5a/scripts/setup/setup.go#L91

---

Fixes #694 **required key GUEST_ADDR missing value**
This issue is caused by k8s version upgrade. After upgrade, they no longer support how we used to configure k8s before. This also fixes #899
https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/kubelet-integration/#the-kubelet-drop-in-file-for-systemd
https://github.com/kubernetes/release/issues/3392

### DIFF
--- a/scripts/cluster/create_multinode_cluster.go
+++ b/scripts/cluster/create_multinode_cluster.go
@@ -74,11 +74,15 @@ func CreateMultinodeCluster(stockContainerd string) error {
 // Create kubelet service on master node
 func CreateMasterKubeletService() error {
 	utils.WaitPrintf("Creating kubelet service")
-	bashCmd := `sudo sh -c 'cat <<EOF > /usr/lib/systemd/system/kubelet.service.d/0-containerd.conf
-[Service]
-Environment="KUBELET_EXTRA_ARGS=--container-runtime=remote --v=%d --runtime-request-timeout=15m --container-runtime-endpoint=unix:///run/containerd/containerd.sock"
+	// Create service directory if not exist
+	_, err := utils.ExecShellCmd("sudo mkdir -p /etc/sysconfig")
+	if !utils.CheckErrorWithMsg(err, "Failed to create kubelet service!\n") {
+		return err
+	}
+	bashCmd := `sudo sh -c 'cat <<EOF > /etc/sysconfig/kubelet
+KUBELET_EXTRA_ARGS="--container-runtime=remote --v=%d --runtime-request-timeout=15m --container-runtime-endpoint=unix:///run/containerd/containerd.sock"
 EOF'`
-	_, err := utils.ExecShellCmd(bashCmd, configs.System.LogVerbosity)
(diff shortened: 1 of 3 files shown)


scripts/utils/system.go
```diff
@@ -183,6 +183,20 @@ func CopyToDir(source string, target string, privileged bool) error {
 	return err
 }
 
+func CopyKernelToDirAndRename(source string, privileged bool) error {
```
ustiugov, line 186: I think this function is quite trivial, please inline it

ustiugov (changes requested):
great! 

Please add "Fixes #936" to the PR description so it closes the issue

see the minor comment inline

P.S. for vHive's rather important changes, please ask me for a review too.

JooyoungPark73:
Ah I missed i opened an issue already. Yes.

JooyoungPark73:
@leokondrashov Can you run them by yourself and check if it works properly?

leokondrashov (approved):
I've checked both single and multi-node setups; all work fine.

ustiugov (approved):

PR #939, opened 2024-02-17
Power Manager 

HermioneKT:
## Summary

This PR uses k8 Power Manager to tune CPU frequency and run experiments (workload sensitivity and internode scaling) to investigate how tuning frequencies can help to optimise power consumption in serverless computing

##  Implementation Notes ⚒️
set up K8 power manager (power profiles, power workloads)
run the experiments using sleeping, spinning, auth, aes benchmarks from vSwarm
A small summary of the requirements (in one/two sentences).

* 

* 


leokondrashov (commented):
I think it is better to place the files differently. Experiment files should be moved to examples, init script to scripts (better to rewrite it inside the setup_tool, but that will be time consuming), leave the library to use in go files in this directory.

PR #940, opened 2024-02-19
K8s control plane high-availability mode

cvetkovic:
## Summary

A small summary of the requirements (in one/two sentences).

* Briefly outline the overall technical solution. If necessary, identify talking points where the reviewer's attention should be drawn to.

* 

* 


leokondrashov:
> NOTE: Only merge this PR in conjunction with https://github.com/vhive-serverless/invitro/pull/379.

I think we need to introduce versioning of vHive in Invitro repo so that we won't need this note. Because it is not only about this PR but also the ongoing knative and k8s upgrade in vHive.

scripts/cluster/create_multinode_cluster.go
```diff
@@ -71,43 +73,22 @@ func CreateMultinodeCluster(stockContainerd string) error {
 	return nil
 }
 
-// Create kubelet service on master node
-func CreateMasterKubeletService() error {
```
leokondrashov, line 75: Don't we need need it? Does it automatically attach to `unix:///run/containerd/containerd.sock`? That's the most important part here. But also, we are loosing the timeout and verbosity parameters. Do we need them?

scripts/cluster/create_multinode_cluster.go
```diff
@@ -71,43 +73,22 @@ func CreateMultinodeCluster(stockContainerd string) error {
...
-func DeployKubernetes() error {
-
+func DeployKubernetes(haReplicaCount int) error {
 	utils.WaitPrintf("Deploying Kubernetes(version %s)", configs.Kube.K8sVersion)
-	masterNodeIp, iperr := utils.ExecShellCmd(`ip route | awk '{print $(NF)}' | awk '/^10\..*/'`)
-	if iperr != nil {
-		return iperr
-	}
-	shellCmd := fmt.Sprintf(`sudo kubeadm init --v=%d \
---apiserver-advertise-address=%s \
```
leokondrashov, line 106: We are loosing it in the default case, does it do it correctly by default?

scripts/cluster/create_multinode_cluster.go
```diff
@@ -172,14 +166,29 @@ func ExtractMasterNodeInfo() error {
...
+		configs.Kube.ApiserverDiscoveryToken,
+		configs.Kube.ApiserverCertificateKey)
 	_, err = masterKeyYamlFile.WriteString(masterKeyYaml)
 	if !utils.CheckErrorWithTagAndMsg(err, "Failed to create masterKey.yaml with master node information!\n") {
 		return err
 	}
 
-	utils.SuccessPrintf("Join cluster from worker nodes with command: sudo kubeadm join %s:%s --token %s --discovery-token-ca-cert-hash %s\n",
-		configs.Kube.ApiserverAdvertiseAddress, configs.Kube.ApiserverPort, configs.Kube.ApiserverToken, configs.Kube.ApiserverTokenHash)
+	utils.SuccessPrintf("Join cluster from worker nodes as a new control plane node with command: "+
```
leokondrashov, line 176: Better to state that you need to join the cluster from backup control plane nodes, not the worker nodes.

scripts/setup.go
```diff
@@ -175,18 +180,18 @@ func main() {
 		err = cluster.SetupWorkerKubelet(setupFlags.Args()[1])
 		// Original scripts from `scripts/cloudlab` directory
 	case "setup_node":
-		if setupFlags.NArg() < 2 {
-			utils.FatalPrintf("Missing parameters: %s <sandbox> [use-stargz]\n", subCmd)
+		if setupFlags.NArg() < 3 {
+			// ha_mode - REGULAR (do not install load balancers), MASTER, BACKUP (with load balancers)
+			utils.FatalPrintf("Missing parameters: %s <ha_mode> <sandbox> [use-stargz]\n", subCmd)
```
leokondrashov, line 185: We need to update the commands in the documentation with these parameters. And better to add the description of HA mode there.

configs/k8s_ha/haproxy.cfg
```diff
@@ -0,0 +1,53 @@
...
+    option httpchk GET /healthz
+    http-check expect status 200
+    mode tcp
+    option ssl-hello-chk
+    balance     roundrobin
+        server control_plane_1 10.0.1.1:6443 check
+        server control_plane_2 10.0.1.2:6443 check
+        server control_plane_3 10.0.1.3:6443 check
+        server control_plane_4 10.0.1.4:6443 check
+        server control_plane_5 10.0.1.5:6443 check
```
leokondrashov, line 53: Does it work with less than five control plane nodes?

leokondrashov (commented):
Pretty good. I left some questions below about the changes, most are just to confirm that it works as expected (I didn't get deep into the HA setting documentation). The most important part is to document these settings in the qiuckstart or place it in separate .md file with links from quickstart. Also, add line into changelog about it; it seems important enough.

configs/k8s_ha/haproxy.cfg
cvetkovic: Yes. It does. The control planes for which it figures out that are not alive are not used as load balancing endpoints.

scripts/cluster/create_multinode_cluster.go
cvetkovic: If you mean the api-server-advertise address, the cluster can work without that. I have tested it both with 1 CP and 3 CPs.

scripts/cluster/create_multinode_cluster.go
cvetkovic: Because `kubeadm` starts `kubelet` process automatically, hence there is no need to start `kubelet` manually. Regarding the `containerd` socket, this is now provided as argument to `kubeadm init` command.

image

@leokondrashov: What is the socket path for Firecracker? Should we make this be configurable?

configs/setup/kube.json
```diff
@@ -6,5 +6,8 @@
 	"PodNetworkAddonConfigURL": "https://raw.githubusercontent.com/vhive-serverless/vHive/main/configs/calico/canal.yaml",
 	"ApiserverPort": "6443",
 	"ApiserverToken": "",
-	"ApiserverTokenHash": ""
+	"ApiserverDiscoveryToken": "",
+	"ApiserverCertificateKey": "",
+	"CPHAEndpoint": "10.0.1.254",
```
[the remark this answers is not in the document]
cvetkovic: That script executes on a regular basis, as a heartbeat in the keepalived. I wouldn't have `jq` on the critical path for performance reasons.

scripts/cluster/create_multinode_cluster.go
leokondrashov: This function doesn't create the kubelet, but rather sets it up with parameters. The socket on the master node doesn't need to be configurable since we use Firecracker only for user containers. But the verbosity and runtime-request-timeout set inside might be used for something.

cvetkovic: Hmm. You are right. 

For the `runtime-request-timeout` parameter, it seems to me we can safely remove this (see [DOCS](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/)). This should mean that each request from the control plane to the kubelet should be executed within 2 minutes (default), rather than 15 minutes.

The `container-runtime` has been deprecated since v1.24 according to [this](https://github.com/cri-o/cri-o/blob/main/tutorials/kubernetes.md).

scripts/README.md
```diff
@@ -1,87 +1,116 @@
 # vHive Setup Scripts
+
 - [vHive Setup Scripts](#vhive-setup-scripts)
```
leokondrashov, line 3: We can remove this syllabus completely, since github has automatic one.

scripts/README.md
```diff
@@ -94,14 +123,29 @@ scripts/gpu/setup_nvidia_gpu.sh
 ./setup_tool [options] setup_nvidia_gpu
 ```
 
-**NOTICE**: Shell scripts in `scripts/stargz`, `scripts/self-hosted-kind`, and `scripts/github_runner` **are not supported to be invoked in this way at present**.
+**NOTICE**: Shell scripts in `scripts/stargz`, `scripts/self-hosted-kind`, and `scripts/github_runner` **are not
+supported to be invoked in this way at present**.
+
+### 3.6 Kubernetes Control Plane High-Availability Mode
 
+For fault tolerance purposes, Kubernetes control plane components can be replicated. This can be done by combining
```
leokondrashov, line 131: Still better have a description of what and how options are used (specifically, REGULAR, which appears throughout the docs).

leokondrashov:
@cvetkovic Hi, what's the status of this PR? My point about the documentation still stands. We should have instructions on how to use the tools in this repo, not just references to general k8s help and a way to use it from Invitro. For example, what would be the steps to setup the HA control plane without all the other things we have in Invitro?

leokondrashov (changes requested):

scripts/cluster/create_multinode_cluster.go
```diff
@@ -141,24 +145,37 @@ func KubectlForNonRoot() error {
...
+	if !utils.CheckErrorWithMsg(err, "Failed to extract API Server address, port, and token from logs!\n") {
 		return err
 	}
 	splittedOut := strings.Split(shellOut, " ")
 	configs.Kube.ApiserverAdvertiseAddress = splittedOut[0]
 	configs.Kube.ApiserverPort = splittedOut[1]
 	configs.Kube.ApiserverToken = splittedOut[2]
+
+	// API Server discovery token
 	shellOut, err = utils.ExecShellCmd("sed -n '/.*sha256:.*/p' < %s/masterNodeInfo | sed -n 's/.*\\(sha256:\\S*\\).*/\\1/p'", configs.System.TmpDir)
```
leokondrashov, line 160: These commands parsing `tmp/masterNodeInfo` return double lines of same info because in the output of `kubeadm init` there are two lines with `kubeadm join` (for control plane and workers). Because of that, masterKey.yaml is malformed.

lrq619:
Hello, what is the status of this PR? Can we merge it now? @leokondrashov

PR #982, opened 2024-05-08
Fix GetNodeIP to get the first ip

lrq619:
## Summary

Change the bash command for getting node's private ip address
Now it would get the first ip address starting with 10.

Change bash command implementation to:
```
ip route | awk '{print $(NF)}' | awk '/^10\..*/ {print; exit}'
```

* N/A

* N/A

### DIFF
--- a/scripts/utils/utils.go
+++ b/scripts/utils/utils.go
@@ -174,8 +174,8 @@ func InstallYQ() {
 
 func GetNodeIP() (string, error) {
 	// Choose the first ip address starting from 10.
-	nodeIP, err := ExecShellCmd(`ip route | awk '{print $(NF)}' | awk '/^10\..*/ {print; exit}'`)
-	if err == nil {
+	nodeIP, err := ExecShellCmd(`ip route | awk '/ src / {for(i=1; i<=NF; i++) if ($i == "src") print $(i+1)}' | awk '/^10\..*/ {print; exit}'`)
+	if (nodeIP != "") && (err == nil) {
 		return nodeIP, nil
 	}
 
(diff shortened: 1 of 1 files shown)


leokondrashov:
Did you check it for operation in the case of two addresses in the subnet?

We should maybe try to let the user choose the IP out of the available ones, as was proposed in the issue.

lrq619:
> Did you check it for operation in the case of two addresses in the subnet?
> 
> We should maybe try to let the user choose the IP out of the available ones, as was proposed in the issue.

Tested by replacing `ip route` to `cat route_ip.txt`, and change the contents in the txt file to simulate enviroments with multiple ip starting from 10.

Enabling user choosing NodeIP is a good idea, but I haven't come up with any ideas.
Most naive one would be pop out a prompt and let user input an ip_address, but I don't think adding such interaction operation would be good in a setup script.It would block there if no input is sent, and this setup script sometimes need to be run by automatical workflows like github actions instead of real person)

Do you have any suggestion?

lrq619:
@JooyoungPark73 I think the gvisor runner is failing again, could you please restart it? Thank you

yulinzou:
Solving this [issue](https://github.com/vhive-serverless/vHive/issues/988), added error handling when `ip route | awk '{print $(NF)}' | awk '/^10\..*/ {print; exit}'` failed to extract IP, will fallback to get from `hostname -I`

lrq619:
@leokondrashov Hi, could you please review this branch so we can get this branch merged?

leokondrashov (approved):
The command works correctly on CloudLab and CCDS cloud. Can merge

PR #1079, opened 2025-04-03
Fix repeated allocations of networks with the same func ID

leokondrashov:
## Summary

Repeated requests for the network with the same func ID were creating a leak in the network manager.

* Check for the allocated networks for the same ID, return it instead of allocating a new one.

* N/A

* N/A


JooyoungPark73:
do we need test for this?

leokondrashov:
I think we can merge without those. (Too late to test this part, but we might need to increase the coverage overall at some point to catch the rest of the bugs)

Issue #1087, opened 2025-04-22
Snapshot not committed due to revision mismatch


andre-j3sus:
**Describe the bug**  
In the Firecracker coordinator's `orchCreateSnapshot` method, a snapshot is initialized with the revision (`fi.Revision` at line 203), but committed using the VM ID (`fi.VmID` at line 231):

https://github.com/vhive-serverless/vHive/blob/71228b09089a61a27225a1d7a8f29a4b59a16fb4/cri/firecracker/coordinator.go#L200-L237

This mismatch causes the snapshot manager to fail to find and commit the snapshot, preventing it from being marked as ready. As a result, the snapshot cannot be used in future runs since it is not marked as ready.

**To Reproduce**  
1. Start vHive with snapshots enabled.  
2. Deploy a function.  
3. Wait until the snapshot is taken.

It should print something like:

```
Snapshot for revision helloworld-1 to commit does not exist
```

**Expected behavior**  
The snapshot should be initialized and committed using the same revision identifier so that it can be properly committed and marked as ready.

**Fix**  
Commit the snapshot with the revision (`fi.Revision`).

PR #1088, opened 2025-04-22
Fix snapshot commit

andre-j3sus:
## Summary

Ensure snapshots are initialized and committed using the same revision identifier to allow successful commit and future reuse.

Refer to #1087 for more details.

* Replaced the incorrect `fi.VmID` with `fi.Revision` in the call to `CommitSnapshot` on `orchCreateSnapshot`.
* Added test (`TestOrchCreateSnapshot`) to verify that the snapshot is correctly committed.

* N/A

* N/A

### DIFF
--- a/.github/workflows/cri_tests.yml
+++ b/.github/workflows/cri_tests.yml
@@ -49,6 +49,10 @@ jobs:
       - name: Run vHive CRI tests
         run: source /etc/profile && go clean -testcache && go test ./cri -v -race -cover
 
+      - name: Run sandbox specific tests
+        if: ${{ inputs.sandbox == 'firecracker' }}
+        run: source /etc/profile && go clean -testcache && go test ./cri/${{ inputs.sandbox }} -v -race -cover
+
       - name: Archive log artifacts
         if: ${{ always() }}
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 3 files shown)


leokondrashov:
Thanks, Andre! LGTM, but remove the newline change in powermanager from the PR.

leokondrashov:
Unfortunately, I don't see the test running in any of the workflows. We would need to add it to the the CI. What was the setup to test it locally? Did it require firecracker running to work? Depending on that we would add it to different runners (self-hosted or github native ones).

andre-j3sus:
> Thanks, Andre! LGTM, but remove the newline change in powermanager from the PR.

I think the newline was added by gofmt, but yes I can remove it.

andre-j3sus:
> Unfortunately, I don't see the test running in any of the workflows. We would need to add it to the the CI. What was the setup to test it locally? Did it require firecracker running to work? Depending on that we would add it to different runners (self-hosted or github native ones).

Oh, that's true. I thought it would run automatically with the other firecracker cri tests. Locally I used the setup described here: https://github.com/vhive-serverless/vHive/blob/main/docs/developers_guide.md#running-cri-tests-locally

leokondrashov (approved):
Thanks for adding the tests in the workflow, they work now. Ready to merge

PR #1095, opened 2025-05-07
Add tests for stargz snapshotter

andre-j3sus:
Add integration tests for firecracker containerd with stargz snapshotter


leokondrashov (approved):
LGTM

PR #1107, opened 2025-05-31
Add support for remote Firecracker snapshots

### LINKED ISSUE #823: Remote firecracker snapshots
**Describe the enhancement**
Rather than only using the snapshots available locally on a node, snapshots can also be transferred between nodes to
potentially accelerate cold start times and reduce memory utilization, given that proper mechanisms are in place to
minimize the snapshot network transfer latency. This could be done by storing snapshots in a global storage solution
such as [MinIO S3](./developers_guide.md#MinIO-S3-service), or directly distributing snapshots between compute nodes.

andre-j3sus:
## Summary

When remote snapshots are enabled, after committing a snapshot, it is uploaded to a MinIO instance. When loading from a snapshot, if it is not available locally, the system checks whether it is available in MinIO and fetches it if found.

Closes #823

## Implementation Notes \:hammer\_and\_pick:

* Changed the `-snapshots` option type from a boolean to a string, with three valid values: `local`, `remote`, or `disabled` (default).
* Added parameters to configure the MinIO connection, including the address, access key, secret key, and bucket name for storing snapshots.
* Implemented a new generic component called `ObjectStorage` with methods to upload and download objects. Created `MinioStorage` as an implementation of this interface.
* Added methods to `SnapshotManager` to support uploading and downloading snapshots via the `ObjectStorage` interface.
* Updated the Firecracker snapshots documentation.
* Added tests for the new components. A MinIO instance is launched as a service in GitHub Actions.

Remote Firecracker snapshots are currently only supported using the Stargz snapshotter (there are known container corruption issues when using devmapper).

## External Dependencies \:four\_leaf\_clover:

* Added `github.com/minio/minio-go/v7` dependency.

## Breaking API Changes \:warning:

* N/A

### DIFF
--- a/cri/firecracker/coordinator.go
+++ b/cri/firecracker/coordinator.go
@@ -139,6 +139,13 @@ func (c *coordinator) stopVM(ctx context.Context, containerID string) error {
 		}
 	}
 
+	if c.orch != nil && c.orch.GetSnapshotMode() == "remote" && fi.SnapBooted && !c.orch.GetCacheSnaps() {
+		if err := c.snapshotManager.DeleteSnapshot(fi.Revision); err != nil {
+			fi.Logger.WithError(err).Error("failed to delete snapshot")
+			return err
+		}
+	}
+
 	return c.orchStopVM(ctx, fi)
 }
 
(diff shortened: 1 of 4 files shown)


andre-j3sus:
The target branch of these PR is currently [stargz-snapshotter](https://github.com/vhive-serverless/vHive/tree/stargz-snapshotter), but will be `main` once #1105 is merged

andre-j3sus:
@leokondrashov I'll temporarily change the target of this PR to avoid triggering CI

PR #1112, opened 2025-06-13
Convert stargz tests to manual tests

andre-j3sus:
## Summary

The unit tests for the Firecracker-containerd interface using the stargz snapshotter were too flaky. It's better to run them in a more controlled way. By running them manually, we can clean and restart Firecracker, making the tests less flaky.

- Changed tests in `ctriface/manual_cleanup_test.go` to allow using stargz snapshotter, by using the same parameters used in `ctriface/iface_test.go`.

N/A

N/A


leokondrashov (commented):
So, you reduced the number of tests that are running, right? Can you compile a list of tests that need fixing into an issue so we can track them? Right now, it feels like back and forth of adding and removing them.

andre-j3sus:
> So, you reduced the number of tests that are running, right? Could you please compile a list of tests that need fixing into an issue so we can track them? Right now, it feels like back and forth of adding and removing them.

@leokondrashov https://github.com/vhive-serverless/vHive/issues/1114

PR #1221, opened 2026-07-09
Integrate Firecracker UFFD handling for UPF snapshots

Liquidzk:
## Summary

Properly integrate Firecracker's UFFD-backed UPF snapshot restore path. Related to #807. 
This PR restores the basic `-snapshots -upf` flow for the current Firecracker stack by wiring Firecracker's vanilla UFFD memory backend into vHive's memory manager.
REAP working-set recording/replay and lazy working-set prefetch need follow-up work.

  * Firecracker updated to 1.13.1
  * Configure UPF restore to use Firecracker `Uffd` memory backend.
  * Receive mappings JSON and one UFFD fd from Firecracker's Unix socket payload.
  * Store mappings in `SnapshotState`.
  * Serve page faults using mapping-based offsets:

    `region.Offset + (faultPageAddr - region.BaseHostVirtAddr)`

  * Remove the old legacy working-set/trace path from this basic implementation.
  * Add memory manager unit tests and CI coverage, including `-race`.

* Firecracker-containerd binaries are based on vhive-v1.8.3 branch in our fork of firecracker-containerd
* Firecracker and jailer binaries are based on v1.13.1-vhive-integration in our fork of firecracker

The existing public flags/options are preserved. `-lazy` / `WithLazyMode` remain accepted for compatibility, but working-set/lazy replay have not been implemented in this PR.


.github/workflows/unit_tests.yml
```diff
@@ -20,6 +20,26 @@ env:
   GO111MODULE: on
 
 jobs:
+  memory-manager-unit-test:
```
leokondrashov: can you combine that with the rest of the unit tests in the matrix?

ctriface/orch.go
```diff
@@ -100,7 +100,6 @@ type Orchestrator struct {
 	// store *skv.KVStore
 	snapshotsEnabled bool
 	isUPFEnabled     bool
-	isLazyMode       bool
```
leokondrashov, line 103: i wouldn't remove the parts that are not working right now like non-lazy mode. just make a condition that it must be lazy (meaning no working set file)

bin/firecracker
```diff
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:23a718b84dbe1a2b0978889657aa6fcb168819426a8603f1a09e4cf7e2a7c425
```
leokondrashov, line 2: it seems that you still use the latest version from my branch where i updated firecracker to 1.13. please explain if the current versions are not satisfactory, because i think it should be fine to use those.

.github/workflows/linters.yml
```diff
@@ -22,6 +22,7 @@ jobs:
       - uses: actions/checkout@v7
         with:
           fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
```
leokondrashov, line 25: is it really needed? it is not used anywhere else

ctriface/iface.go
```diff
@@ -583,51 +604,56 @@ func (o *Orchestrator) LoadSnapshot(ctx context.Context, vmID string, snap *snap
...
-				snapFiles += f.Name() + ", "
-			}
-			logger.Error(snapFiles)
+			logSnapshotLoadFailure(logger, snap, conf, loadErr)
 		}
-	}()
-
-	if o.GetUPFEnabled() {
-		if activateErr = o.memoryManager.Activate(vmID); activateErr != nil {
+		activateErr = <-activateDone
```
leokondrashov: can you move this activate error handling into the goroutine at line 631? this way, we would have an if for upf enabled and common part of create vm that runs always.

ctriface/iface.go
```diff
@@ -583,51 +604,56 @@ func (o *Orchestrator) LoadSnapshot(ctx context.Context, vmID string, snap *snap
...
+		go func() {
+			defer close(loadDone)
+
+			if _, loadErr = o.fcClient.CreateVM(ctx, conf); loadErr != nil {
+				logSnapshotLoadFailure(logger, snap, conf, loadErr)
+			}
+		}()
 
-	<-loadDone
+		<-loadDone
```
leokondrashov: this doesn't make any sense now, we create a goroutine and wait right after it. it shouldn't be a goroutine anymore. previously it was needed because of activation process that required us to launch the creation and activation in parallel. now we should do the opposite.

memory/manager/manager.go
```diff
@@ -117,14 +143,24 @@ func (m *MemoryManager) DeregisterVM(vmID string) error {
 
 // Activate Creates an epoller to serve page faults for the VM
 func (m *MemoryManager) Activate(vmID string) error {
+	return m.activate(vmID, nil)
+}
+
+// ActivateWithSocketReady is like Activate, but reports when the UFFD socket
+// listener is ready for Firecracker to connect.
```
leokondrashov: if we need this signal to work properly, why do we have a version that doesn't send it?

.github/workflows/unit_tests.yml
```diff
@@ -64,8 +64,12 @@ jobs:
           AWS_ACCESS_KEY: ${{ secrets.AWS_ACCESS_KEY }}
           AWS_SECRET_KEY: ${{ secrets.AWS_SECRET_KEY }}
       run: |
-        make -C $MODULE test
-        make -C $MODULE test-man
+        if [[ "$MODULE" == "memory/manager" ]]; then
+          go test -race ./memory/manager -count=1
```
leokondrashov: can you create the makefile so it would run the same way as the rest of the modules

ctriface/iface_test.go
```diff
@@ -51,14 +51,19 @@ var (
...
 
-	if *isUPFEnabled {
-		log.Error("User-level page faults are temporarily disabled (gh-807)")
-		os.Exit(-1)
-	}
-
 	os.Exit(m.Run())
 }
 
+func TestValidateUPFModeRequiresLazyMode(t *testing.T) {
```
leokondrashov: don't think there is a need for this test. this is temporary arrangement with disabled non-lazy mode

memory/manager/manager.go
```diff
@@ -133,46 +164,62 @@ func (m *MemoryManager) Activate(vmID string) error {
 	if !ok {
 		m.Unlock()
 		logger.Error("VM not registered with the memory manager")
-		return errors.New("VM not registered with the memory manager")
+		err := errors.New("VM not registered with the memory manager")
+		notifySocketReady(socketReadyCh, err)
+		return err
```
leokondrashov: this seems very weird; we are reporting the error twice. and in iface.go, that leads to both of them collated into a multierror. this is for sure redundant.

ctriface/iface_test.go
leokondrashov, line 1: for proper tests, can you enable the tests with upfs in the makefile? those are the most important for us to understand whether it works or not. obviously, you need to handle the cases of non-lazy mode since it is not available now.

leokondrashov (commented):
please prioritize enabling iface tests with upfs. those are the ones i would need operational to accept the pr.

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/vhive-serverless/vhive
 go 1.26.2
 
 replace (
-	github.com/firecracker-microvm/firecracker-containerd => github.com/vhive-serverless/firecracker-containerd v0.0.0-20251220120737-a3dabeb332c7
+	github.com/firecracker-microvm/firecracker-containerd => github.com/vhive-serverless/firecracker-containerd v0.0.0-20260205055657-a35b52c279d9
```
leokondrashov:
```suggestion
	github.com/firecracker-microvm/firecracker-containerd => github.com/vhive-serverless/firecracker-containerd v0.0.0-20260713114451-bb007a1ef569
```

this is the branch i used to generate the binaries. it shouldn't create any visible differences, but still better to refer the most up-to-date

ctriface/iface.go
```diff
@@ -583,51 +610,53 @@ func (o *Orchestrator) LoadSnapshot(ctx context.Context, vmID string, snap *snap
...
-			snapFiles := ""
-			for _, f := range files {
-				snapFiles += f.Name() + ", "
-			}
-
-			logger.Error(snapFiles)
-
-			files, _ = os.ReadDir(filepath.Dir(conf.ContainerSnapshotPath))
+	if o.GetUPFEnabled() {
+		activateDone = make(chan error, 1)
```
leokondrashov: please rename to activateErrChan or something similar, it is confusing that we have activate done and soket ready signals that mean completely different things.

memory/manager/snapshot_state.go
```diff
@@ -49,6 +50,159 @@ import (
...
+
+	regionOffset := faultPageAddr - region.BaseHostVirtAddr
+	if region.Offset > math.MaxUint64-regionOffset {
+		return 0, fmt.Errorf("guest memory offset overflow for fault address %#x", faultPageAddr)
+	}
+
+	return region.Offset + regionOffset, nil
+}
+
+func guestAddressForMemoryOffset(region GuestRegionUffdMapping, offset uint64) (uint64, error) {
```
leokondrashov: a lot of functions here are not about snapshot but upf utilities. please extract those into a separate file. and add some short docs for each, for some of those functions, it's hard to get what is the goal of the function.

memory/manager/snapshot_state.go
```diff
@@ -202,88 +534,93 @@ func (s *SnapshotState) unmapGuestMemory() error {
...
+	return nil
+}
+
+// alignment returns alignment of the block in memory with reference to alignSize.
+func alignment(block []byte, alignSize int) int {
+	return int(uintptr(unsafe.Pointer(&block[0])) & uintptr(alignSize-1))
+}
+
+// AlignedBlock returns a []byte aligned to the host page size.
+func AlignedBlock(blockSize int) []byte {
```
leokondrashov: those functions look a lot like the ones in the beginning of the file. do we need both of them? i don't think we need to have several implementations of the same alignment logics

ctriface/iface.go
```diff
@@ -583,51 +610,53 @@ func (o *Orchestrator) LoadSnapshot(ctx context.Context, vmID string, snap *snap
...
 		}
-	}()
+	}
 
-	if o.GetUPFEnabled() {
-		if activateErr = o.memoryManager.Activate(vmID); activateErr != nil {
-			logger.Warn("Failed to activate VM in the memory manager", activateErr)
-		}
+	if _, loadErr = o.fcClient.CreateVM(ctx, conf); loadErr != nil {
+		logSnapshotLoadFailure(logger, snap, conf, loadErr)
```
leokondrashov, line 656: when it fails here, we don't deactivate the vm, that is a leak

leokondrashov (commented):
please work on the following. there are issues with the code structure, especially in the snapshot_state.go.

leokondrashov (approved):
now looks fine to me. please work on the working sets now

PR #1229, opened 2026-07-22
UPF working set restoring

Liquidzk:
## Summary

  Follow-up to https://github.com/vhive-serverless/vHive/pull/1221. This PR restores UPF working-set recording and replay.

  - Stores working-set pages and traces as snapshot artifacts.
  - Installs the recorded working set during subsequent snapshot restores.
  - Adds tests for recording and replaying working sets across different VM instances.

The legacy working-set design relied on VM-local in-memory state and did not fit revision-based restores. (exists some differences from the old version) This implementation therefore:

- Stores artifacts by revision snapshot instead of VM ID.
- Persists versioned trace metadata for reuse across VM instances.
- Uses Firecracker mappings and page sizes for address translation.
- Validates and atomically replaces working-set artifacts.

* 

* 

### DIFF
--- a/ctriface/Makefile
+++ b/ctriface/Makefile
@@ -24,6 +24,7 @@ EXTRAGOARGS:=-v -race -cover
 EXTRATESTFILES:=iface_test.go iface.go orch_options.go orch.go
 BENCHFILES:=bench_test.go iface.go orch_options.go orch.go
 UPFARGS:=-upf -lazy
+REAPARGS:=-upf
 STARGZ:=-ss 'proxy' -img 'ghcr.io/vhive-serverless/helloworld:var_workload-esgz'
 DOCKER_CREDENTIALS:=-dockerCredentials '{"docker-credentials":{"ghcr.io":{"username":"","password":""}}}'
 GOBENCH:=-v -timeout 1500s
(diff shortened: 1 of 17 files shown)


vhive.go
leokondrashov, line 1: please enable the tests in the main Makefile. look for any other references to 807 in the codebase to catch all remnants of the disabled upfs. including docs.

ctriface/Makefile
```diff
@@ -37,6 +38,11 @@ test:
 	sudo mkdir -m777 -p $(CTRDLOGDIR) && sudo env "PATH=$(PATH)" /usr/local/bin/firecracker-containerd --config /etc/firecracker-containerd/config.toml 1>$(CTRDLOGDIR)/ctriface_log.out 2>$(CTRDLOGDIR)/ctriface_log.err &
 	sudo env "PATH=$(PATH)" go test $(EXTRATESTFILES) $(EXTRAGOARGS) -args $(UPFARGS)
 	./../scripts/clean_fcctr.sh
+	@set -e; \
+	trap './../scripts/clean_fcctr.sh' EXIT; \
+	sudo mkdir -m777 -p $(CTRDLOGDIR); \
```
leokondrashov: please make it like the previous tests: only the clean_fcctr. directory is already created and set -e is not needed

ctriface/orch_options.go
```diff
@@ -73,8 +73,8 @@ func WithLazyMode(isLazyMode bool) OrchestratorOption {
 }
 
 func (o *Orchestrator) validateUPFMode() error {
```
leokondrashov, line 75: i missed it last time, but can you move the call of validateUPFMode in to orchestrator creation, rather than load snapshot. I would prefer this validation fail on initialization of vhive rather than on the first vm restore.

memory/manager/trace.go
```diff
@@ -49,6 +50,12 @@ type Trace struct {
 	regions          map[uint64]int
 }
 
+type workingSetTraceMetadata struct {
```
leokondrashov: i would call it differently, it is not metadata, it is a serialized trace format, so i would call it `seralizedWorkingSet`

leokondrashov (commented):
looks good, please fix the minor comments

leokondrashov (approved):
looks good now

